PR #211, opened 2024-06-12
Formatting fixes to e2e tests

RHARV018:
## Type of change 🧩

- [ ] Documentation update
- [x] Other (please explain in the description section above)

### DIFF
--- a/appeals/e2e/cypress/e2e/back-office-appeals/reviewLpaq.spec.js
+++ b/appeals/e2e/cypress/e2e/back-office-appeals/reviewLpaq.spec.js
@@ -7,14 +7,14 @@ import { UpdateDueDatePage } from '../../page_objects/updateDueDatePage';
 const appealsListPage = new AppealsListPage();
 const updateDueDatePage = new UpdateDueDatePage();
 
-describe.skip('Appeals feature', () => {
+describe('Appeals feature', () => {
 	beforeEach(() => {
 		cy.login(users.appeals.caseAdmin);
 	});
 
 	it('Complete LPAQ', () => {
-		cy.visit('/appeals-service/appeals-list');
-		appealsListPage.clickAppealFromList(18);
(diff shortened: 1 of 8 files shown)


appeals/e2e/cypress/e2e/back-office-appeals/reviewLpaq.spec.js
```diff
@@ -7,14 +7,14 @@ import { UpdateDueDatePage } from '../../page_objects/updateDueDatePage';
...
+describe('Appeals feature', () => {
 	beforeEach(() => {
 		cy.login(users.appeals.caseAdmin);
 	});
 
 	it('Complete LPAQ', () => {
-		cy.visit('/appeals-service/appeals-list');
-		appealsListPage.clickAppealFromList(18);
+		cy.visit('/appeals-service/all-cases');
+		appealsListPage.clickAppealFromList(30);
```
PeterBridgerKC, line 17: This approach feels fragile, as the appeal we want has already changed in position. I suggest that we revisit this and select the appeal based on data attribute, so that the position of it in the list doesn't matter

PeterBridgerKC:
From a perspective of these changes getting the tests working again, I'm happy to approve

However the next here is to undertake the work in BOAT-1198/1148 to add `data-cy` attributes to the key page elements so that we don't need to rely on position based selection methods. As can be seen in some of the changes in this PR, quite a few of the fixes are due to the position of an element changing (e.g. position in a list)

PeterBridgerKC (approved):

appeals/e2e/cypress/e2e/back-office-appeals/reviewLpaq.spec.js
RHARV018: I agree, once we have the data attribute we can refine all of our tests that use positions, element ids or text copy to make the more stable.

PR #227, opened 2024-06-17
fix(api): fix address update to allow removal of data from optional fields

AlexMoulin:
Was unable to clear data when it was no longer required, so changing addressLine2 from a string to an empty is now possible.

## Type of change 🧩

- [ ] Documentation update

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/appeals/api/src/server/common/validators/string-parameter.js
+++ b/appeals/api/src/server/common/validators/string-parameter.js
@@ -31,7 +31,7 @@ const validateStringParameter = (parameterName, maxLength = LENGTH_300) =>
  */
 export const validateStringParameterAllowingEmpty = (parameterName, maxLength = LENGTH_300) =>
 	body(parameterName)
-		.optional()
+		.optional({ nullable: true })
 		.isString()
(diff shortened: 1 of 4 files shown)


appeals/api/src/server/common/validators/string-parameter.js
```diff
@@ -31,7 +31,7 @@ const validateStringParameter = (parameterName, maxLength = LENGTH_300) =>
  */
 export const validateStringParameterAllowingEmpty = (parameterName, maxLength = LENGTH_300) =>
 	body(parameterName)
-		.optional()
+		.optional({ nullable: true })
```
glpins, line 34: not sure if this was necessary ?

glpins (approved):

appeals/api/src/server/common/validators/string-parameter.js
AlexMoulin: It was unfortunately as it was submitting a null which optional doesn't accept by default.

glpins (approved):

PR #231, opened 2024-06-19
fix(appeals): correct avscan for documents not yet committed

fabfab-kc:
As a result of AVI-80, and the changes in the document upload process, the AV scan was failing to report results on documents that were added to blob storage, but not yet committed to the database.

To fix, the DocumentVersionAvScan table (were AV scans are written), no longer has a constraint to the DocumentVersion table, allowing scan reports to be stored independently.

Fix to [AVI-80](https://pins-ds.atlassian.net/browse/ASI-80)

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/src/server/endpoints/documents/documents.controller.js
+++ b/appeals/api/src/server/endpoints/documents/documents.controller.js
@@ -316,17 +316,18 @@ const updateDocumentsAvCheckStatus = async (req, res) => {
 	const { body } = req;
 	try {
 		const documents = body.documents;
+		await documentRepository.createDocumentAvStatus(documents);
+
 		for (const document of documents) {
-			const latestDocument = await documentRepository.getDocumentWithAllVersionsById(document.id);
-			if (!latestDocument) {
-				return res.status(404).send({ errors: { body: ERROR_NOT_FOUND } });
-			}
-			const versionList = latestDocument.versions?.map((v) => v.version);
-			if (!versionList || versionList.indexOf(document.version) < 0) {
-				return res.status(404).send({ errors: { body: ERROR_NOT_FOUND } });
+			const latestDocument = await documentRepository.getDocumentByIdAndVersion(
+				document.id,
+				document.version
+			);
+			const versionList = latestDocument?.versions?.map((v) => v.version);
+			if (versionList && versionList.indexOf(document.version) > -1) {
+				await documentRepository.updateDocumentAvStatus(document);
 			}
 		}
-		await documentRepository.updateDocumentAvStatus(documents);
 	} catch (error) {
 		if (error) {
 			logger.error(error);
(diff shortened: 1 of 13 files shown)


appeals/api/src/server/endpoints/documents/documents.controller.js
```diff
@@ -316,17 +316,15 @@ const updateDocumentsAvCheckStatus = async (req, res) => {
...
-			}
-			const versionList = latestDocument.versions?.map((v) => v.version);
-			if (!versionList || versionList.indexOf(document.version) < 0) {
-				return res.status(404).send({ errors: { body: ERROR_NOT_FOUND } });
+			const versionList = latestDocument?.versions?.map((v) => v.version);
+			if (versionList && versionList.indexOf(document.version) < 0) {
+				await documentRepository.updateDocumentAvStatus(document);
 			}
 		}
-		await documentRepository.updateDocumentAvStatus(documents);
```
beejjacobs, line 329: I can't work out if this should be in a transaction or not. I think as we create the AV status first, this might be OK.

appeals/api/src/server/endpoints/documents/documents.controller.js
```diff
@@ -316,17 +316,15 @@ const updateDocumentsAvCheckStatus = async (req, res) => {
 	const { body } = req;
 	try {
 		const documents = body.documents;
+		await documentRepository.createDocumentAvStatus(documents);
+
 		for (const document of documents) {
 			const latestDocument = await documentRepository.getDocumentWithAllVersionsById(document.id);
```
beejjacobs, line 322: Do we need to fetch all versions here?

beejjacobs (commented):
Generally looks good. Just some questions I am not sure on...

appeals/api/src/server/endpoints/documents/documents.controller.js
fabfab-kc: Yeah we write the AV result in the disconnected table first (line 319), then we want to update the DocumentVersion rows that we can find (to cover records that may already be in the DB), on line 325

appeals/api/src/server/endpoints/documents/documents.controller.js
fabfab-kc: I think we do, as the scan will not necessarily update the latest version

beejjacobs: Can we not just fetch the version we want to update? Or even just call update and ignore not found errors?

appeals/api/src/server/endpoints/documents/documents.controller.js
beejjacobs: I think because `addDocument` uses a transaction when checking for scan status, this is OK.

appeals/api/src/server/repositories/document.repository.js
```diff
@@ -20,14 +20,34 @@ export const getDocumentById = (guid) => {
...
+ * @returns {PrismaPromise<Document|null>}
+ */
+export const getDocumentByIdAndVersion = (guid, version) => {
+	return databaseConnector.document.findUnique({
+		where: { guid },
+		include: {
+			versions: {
+				include: {
+					redactionStatus: true
+				},
```
beejjacobs, line 42: Does this mean it won't update virus scan status unless it's redacted?

fabfab-kc: no, actually that's to bring the redactionStatus data - but is not needed in this call. I'll remove

fabfab-kc: done

beejjacobs: Ah I misunderstood, not a where but an include clause!

beejjacobs (approved):
Comments resolved, LGTM.

PR #254, opened 2024-06-28
feat(api): seeding for env

fabfab-kc:
Seeds LPAs from a different list depending on the environment

## Type of change 🧩

- [ ] Documentation update


appeals/api/src/database/seed/seed-training.js
```diff
@@ -0,0 +1,24 @@
...
+		await seedLPAs(databaseConnector, localPlanningDepartmentList);
+	} catch (error) {
+		console.error(error);
+		throw error;
+	} finally {
+		await databaseConnector.$disconnect();
+	}
+};
+
+await (process.env.NODE_ENV === 'training' ? seedTraining() : null);
```
beejjacobs, line 24: Not sure we need to set/check `NODE_ENV` (I know copied from the prod one, but I wasn't sure why that was there either!

appeals/api/src/database/seed/LPAs/prod.js
```diff
@@ -0,0 +1,12 @@
+/** @typedef {Omit<import('@pins/appeals.api').Schema.LPA, 'id'>} LPA */
```
beejjacobs, line 1: We should use the create Prisma type here, it'll be something like

import('<db-client>').Prisma.LpaCreateInput

appeals/api/src/database/seed/seed-training.js
fabfab-kc: yeah wasn't so sure about that... those are passed to the npm scripts with CROSS-ENV

azure-pipelines-db-seed-appeals.yml
beejjacobs, line 1: `keyVaultName` needs to end in just `train` for the training env

beejjacobs (approved):
LGTM.

Might be worth running the pipeline to check for syntax errors etc...

PR #264, opened 2024-07-02
fix(web): fix status-dependent action link copy and URLs on personal list

james-hill-kc:
## Describe your changes

Fix status-dependent action link copy and URLs on personal list, plus some additional changes:
- changed default action for personal list from ‘View appellant case’ to ‘View case’ (URL not changed as it was already linking to case details page, not appellant case)
- amended all personal list actions copy casing to `Sentence case` instead of `Title Case` as requested by content designer
- updated status tag template to account for `lpa_questionnaire_due` status having changed to `lpa_questionnaire`
- update personal list `mapAppealStatusToActionRequiredHtml` function to use STATUSES values from shared constants file, to avoid future regressions if state names change
- update unit tests in WEB to update `lpa_questionnaire_due` status value to `lpa_questionnaire`

BOAT-1390

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/lib/nunjucks-filters/status-tag.js
+++ b/appeals/web/src/server/lib/nunjucks-filters/status-tag.js
@@ -10,7 +10,7 @@ export function appealStatusToStatusTag(appealStatus) {
 	return capitalizeFirstLetter(
 		appealStatus
 			.replace('issue_determination', 'issue_decision')
-			.replace('lpa_questionnaire_due', 'lpa_questionnaire')
+			.replace('lpa_questionnaire_due', 'lpa_questionnaire') // TODO: remove once status value updates are complete
 			.replace('lpa_', 'LPA_')
 			.replace('lpaq_', 'LPAQ_')
 			.replaceAll('_', ' ')
--- a/appeals/web/src/server/views/appeals/components/status-tag.njk
+++ b/appeals/web/src/server/views/appeals/components/status-tag.njk
@@ -9,7 +9,9 @@
 	{%- switch (status | string | lower) -%}
 		{%- case 'ready_to_start' -%}
 			{%- set statusTagColor = 'turquoise' -%}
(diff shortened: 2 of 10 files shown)


appeals/web/src/server/views/appeals/components/status-tag.njk
```diff
@@ -9,7 +9,7 @@
 	{%- switch (status | string | lower) -%}
 		{%- case 'ready_to_start' -%}
 			{%- set statusTagColor = 'turquoise' -%}
-		{%- case 'lpa_questionnaire_due' -%}
```
fabfab-kc, line 12: I think we should keep this until we align the API, otherwise will fail as real appeals will still have `lpa_questionnaire_due`

appeals/web/src/server/lib/nunjucks-filters/status-tag.js
```diff
@@ -10,7 +10,6 @@ export function appealStatusToStatusTag(appealStatus) {
 	return capitalizeFirstLetter(
 		appealStatus
 			.replace('issue_determination', 'issue_decision')
-			.replace('lpa_questionnaire_due', 'lpa_questionnaire')
```
fabfab-kc, line 13: Probably also keep this for now? See below...

fabfab-kc (changes requested):
I think we should do the API work first, or - if a blocker - maybe the suggested changes?

appeals/web/src/server/lib/nunjucks-filters/status-tag.js
james-hill-kc: Done - I've also updated `mapAppealStatusToActionRequiredHtml` to check for the old value as well as the new one, so should handle both until the status can be globally updated

fabfab-kc (approved):
LGTM!

PR #405, opened 2024-08-22
feat(web): add show more component

james-hill-kc:
## Describe your changes

add show more component which hides text in summary list rows when it is longer than 300 characters (only implemented for appellant case > original development description row currently)

A2-763

## Type of change 🧩

- [ ] Documentation update


appeals/web/src/client/components/show-more/show-more.js
```diff
@@ -0,0 +1,106 @@
...
+	const labelText = componentInstance.elements.root.getAttribute(ATTRIBUTES.label);
+
+	let html = '';
+
+	html += `<span class="${CLASSES.text}">${previewText}</span>`;
+	html += `<span class="${CLASSES.ellipsis}" aria-hidden="true">&hellip;</span> `;
+	html += `<button class="${CLASSES.toggleButton}" aria-expanded="false">`;
+	html += `<span class="${CLASSES.toggleButtonLabel}">${DEFAULT_OPTIONS.toggleButtonTextCollapsed}</span>`;
+	html += `<span class="govuk-visually-hidden">, ${labelText}</span>`;
+	html += '</button>';
```
kandcdev, line 82: Minor: can whole HTML be replaced with template literals? ie.

```
const html = `
  <span class="${CLASSES.text}">${previewText}</span>
  <span class="${CLASSES.ellipsis}" aria-hidden="true">&hellip;</span>
  <button class="${CLASSES.toggleButton}" aria-expanded="false">
  <span class="govuk-visually-hidden">, ${labelText}</span>
  </button>
`;

I can be a bit more readable

james-hill-kc: I considered this but thought the line breaks it adds would interfere with the way the content is read by screen readers. But I tried it and it seems to work fine, so I've updated it now

kandcdev (approved):

PR #451, opened 2024-09-10
feat(web): feature/A2-887-no-duplicate-filenames-to-be-uploaded

kandcdev:
## Describe your changes

https://pins-ds.atlassian.net/browse/A2-887

This PR checks the list of filenames in the `currentFolder` and displays an error if the user attempts to upload a file with a name that already exists in that list.

## Type of change 🧩

- [ ] Documentation update

- [] I have made corresponding changes to the documentation

### DIFF
--- a/appeals/web/src/server/appeals/appeal-documents/appeal-documents.controller.js
+++ b/appeals/web/src/server/appeals/appeal-documents/appeal-documents.controller.js
@@ -66,6 +66,17 @@ export const renderDocumentUpload = async (
 		return response.status(404).render('app/404.njk');
 	}
 
+	const filenamesInFolder = currentFolder.documents
+		? Buffer.from(
+				JSON.stringify(
+					currentFolder.documents.map(
+						(/** @type {import('@pins/appeals.api').Appeals.DocumentInfo} */ documentInfo) =>
+							documentInfo.name
+					)
+				)
+		  ).toString('base64')
+		: '';
+
 	let documentName;
 	let _documentType = documentType;
 	let latestVersion;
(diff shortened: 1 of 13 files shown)


appeals/web/src/server/appeals/appeal-documents/appeal-documents.controller.js
```diff
@@ -66,6 +66,17 @@ export const renderDocumentUpload = async (
...
 
+	const filenamesInFolder = currentFolder.documents
+		? Buffer.from(
+				JSON.stringify(
+					currentFolder.documents.map(
+						(/** @type {import('@pins/appeals.api').Appeals.DocumentInfo} */ documentInfo) =>
+							documentInfo.name
+					)
+				)
+		  ).toString('base64')
```
Si3rr4wow, line 77: I might've missed some discussion on this but how come you're base 64 encoding on the server only to decode it on the client to check the names? I also might've misunderstood what the code does

kandcdev: I went with Base64 because I needed to send a JSON object from the server to the client. Base64 makes sure there’s no weirdness with special characters, like double quotes, when you pull it from the "data" HTML attribute. It’s kind of the go-to move for keeping things clean when passing info between the server and client. In this case i am storing a JSON of an array of file names in `data-filenames-in-folder`

Si3rr4wow: Ah, I've never seen this hack before but I've also never been on a gov project where we were allowed clientside js

Si3rr4wow (approved):

PR #432, opened 2024-09-03
Refactor/a2 57 break up mapper

Si3rr4wow:
## Type of change 🧩

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please explain in the description section above)

## Checklist before requesting a review

- [ ] I have performed a self-review of my own code
- [ ] I have double checked this work does not include any hardcoded secrets or passwords
- [ ] I have made corresponding changes to the documentation
- [ ] I have provided details on how I have tested my code
- [ ] I have referenced the ticket number above
- [ ] I have provided a description of the ticket
- [ ] I have included unit tests to cover any testable code changes

PR #492, opened 2024-10-01
Task/page mapper improvements

beejjacobs:
## Describe your changes

Best reviewed one commit at a time! This reduces the copy+pasted code when setting up the mappers for appeals, appellant case, and LPAQ.

There are now functions such as `textDisplayField`, `booleanDisplayField`, `documentTypeDisplayField`.

- [refactor(api): move duplicate document middleware into common folder](https://github.com/Planning-Inspectorate/appeals-back-office/commit/93cb709e24e6416e32fa3a9884e7ce33567ebf04)
- [refactor(web): add default text option to yes/no mapper](https://github.com/Planning-Inspectorate/appeals-back-office/commit/72a46e9f028970420bee8b3020534f84bc2ba152)
- [refactor(web): add boolean display field helper](https://github.com/Planning-Inspectorate/appeals-back-office/commit/f86cee5d1fdf30d8a2435e50a6b47bccaa7ff404)
- [refactor(web): add document display field helper](https://github.com/Planning-Inspectorate/appeals-back-office/commit/36e7dd9bfee7ada3e44225a1e3be146e2519231d)
- [refactor(web): add text display field helper](https://github.com/Planning-Inspectorate/appeals-back-office/commit/74f4fe85d9b447b97a48aae841be9bbb9b5cf0b9)

## Type of change 🧩

- [ ] Documentation update


appeals/web/src/server/lib/mappers/appellantCase.mapper.js
```diff
@@ -29,348 +31,151 @@ export function initialiseAndMapData(appellantCaseData, appealDetails, currentRo
...
+	 * Generates a document field with defaults
+	 *
+	 * @param {Object} options
+	 * @param {string} options.id
+	 * @param {string} options.text
+	 * @param {FolderInfo|null|undefined} options.folderInfo
+	 * @param {string} [options.cypressDataName]
+	 * @returns {Instructions}
+	 */
+	const documentField = ({ id, text, folderInfo, cypressDataName }) => {
```
james-hill-kc, line 44: This name feels a bit vague - the term "field" is too ambiguous in my opinion. Maybe "instruction" would be a better fit?

appeals/web/src/server/lib/page-components/document.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {number} options.appealId
+ * @param {import('@pins/appeals.api').Appeals.FolderInfo|null|undefined} options.folderInfo
+ * @param {string} options.requiredPermissionName
+ * @param {import('../../app/auth/auth-session.service.js').SessionWithAuth} options.session
+ * @param {string} options.manageUrl
+ * @param {string} options.uploadUrlTemplate
+ * @param {string} [options.cypressDataName]
+ * @returns {Instructions}
+ */
+export function documentTypeDisplayField({
```
james-hill-kc, line 20: Again I think `field` is too generic here. This is returning a summary list item display instruction, ideally the name would reflect that

appeals/web/src/server/lib/mappers/lpaQuestionnaire.mapper.js
```diff
@@ -29,6 +32,34 @@ export function initialiseAndMapLPAQData(
...
+	 * Generates a document field with defaults
+	 *
+	 * @param {Object} options
+	 * @param {string} options.id
+	 * @param {string} options.text
+	 * @param {FolderInfo|null|undefined} options.folderInfo
+	 * @param {string} [options.cypressDataName]
+	 * @returns {Instructions}
+	 */
+	const documentField = ({ id, text, folderInfo, cypressDataName }) => {
```
james-hill-kc, line 45: As mentioned elsewhere I think this should be more descriptively named. I'm not sure why the wrapper function is necessary if all it's doing is calling `documentTypeDisplayField` and returning the result. Also there are variables here being referenced from the containing scope - I think those should be passed in as parameters for clarity

appeals/web/src/server/lib/page-components/boolean.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {string} options.id
+ * @param {string} options.text
+ * @param {boolean|null} [options.value]
+ * @param {string} [options.defaultText]
+ * @param {string} options.link
+ * @param {boolean} [options.addCyAttribute]
+ * @param {import("express-session").Session & Partial<import("express-session").SessionData>} options.session
+ * @returns {Instructions}
+ */
+export function booleanDisplayField({
```
james-hill-kc, line 17: Could be more descriptively named, ideally the name would include `instruction` as that's what's being returned

appeals/web/src/server/lib/page-components/boolean.js
```diff
@@ -0,0 +1,49 @@
...
+	id,
+	text,
+	value,
+	defaultText,
+	link,
+	addCyAttribute,
+	session
+}) {
+	return {
```
james-hill-kc, line 26: Would it be more flexible to create a helper function to generate summaryListItems, and have data-type-specific functions which call that, eg. for boolean values in this case?

appeals/web/src/server/lib/page-components/document.js
```diff
@@ -0,0 +1,64 @@
...
+	id,
```
james-hill-kc, line 21: Just FYI there's a tech debt ticket here to refactor the shared document code to use this destructured object pattern for parameters. If it's been decided this approach should be used for all new code, should we extend that to cover all existing functions in the codebase? https://pins-ds.atlassian.net/browse/A2-949

appeals/web/src/server/lib/page-components/document.js
```diff
@@ -0,0 +1,64 @@
...
+		text: 'Add',
+		visuallyHiddenText: text,
+		href: formatDocumentActionLink(appealId, folderInfo, uploadUrlTemplate),
+		attributes: { 'data-cy': `add-${cypressDataName}` }
+	});
+	add && actions.push(add);
+	return {
+		id,
+		display: {
+			summaryListItem: {
```
james-hill-kc, line 53: This would be another potential use case for the proposed summary list item helper

appeals/web/src/server/lib/page-components/text.js
```diff
@@ -0,0 +1,66 @@
...
+import { permissionNames } from '#environment/permissions.js';
+import { SHOW_MORE_MAXIMUM_CHARACTERS_BEFORE_HIDING } from '#lib/constants.js';
+
+/**
+ * Returns the instructions for a text display field
+ *
+ * @param {Object} options
+ * @param {string} options.id
+ * @param {string} options.text
+ * @param {string|{html: string}|null} [options.value]
```
james-hill-kc, line 11: You could potentially use the type `HtmlProperty` defined in `appeals/web/src/server/lib/nunjucks-template-builders/gds-components-properties-jsdoc.js` here, but that does also include an optional `pageComponents` property - just a thought

james-hill-kc (commented):
Looking great overall. In spirit I agree with the changes and it's nice to see duplication being reduced. My only major concern is around naming of functions and variables not being descriptive enough or specific enough. More verbose code is more cumbersome to work with, but it's also easier to understand for someone less familiar with the codebase. When doing this type of refactoring, we risk fixing one problem by creating another worse problem (for me personally, trying to decipher how something works is generally more difficult than wading through duplicated code). It's really important that we preserve as much clarity as possible, and in my view proper naming is critical to that. Apologies if this seems like a pedantic point; I just feel strongly about this topic!

AlexMoulin (dismissed):

appeals/web/src/server/lib/mappers/appellantCase.mapper.js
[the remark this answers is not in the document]
beejjacobs: This is defined here so it can access the session, appeal ID, etc... - these parts are common for each document row, so its just a small wrapper to set those values. It can't be in the top scope as then it wouldn't have access to those values.

beejjacobs: OK, we can go with `documentInstruction`

appeals/web/src/server/lib/page-components/document.js
beejjacobs: Gone with `documentTypeDisplayInstruction`

appeals/web/src/server/lib/mappers/lpaQuestionnaire.mapper.js
beejjacobs: Those parameters are the same for all uses in the mapper, so the idea is to reduce ~4 lines of copied code. With this approach it is clear those parameters are the same for each row.

appeals/web/src/server/lib/page-components/boolean.js
beejjacobs: Gone with `booleanDisplayInstruction`

appeals/web/src/server/lib/page-components/boolean.js
beejjacobs: Possibly, I think the duplicate code between the summary list items is very minimal. Might be worth exploring though.

appeals/web/src/server/lib/page-components/document.js
```diff
@@ -0,0 +1,64 @@
...
+	folderInfo,
+	requiredPermissionName,
+	session,
+	manageUrl,
+	uploadUrlTemplate,
+	cypressDataName = id
+}) {
+	const documents = (isFolderInfo(folderInfo) && folderInfo.documents) || [];
+	/** @type {ActionItemProperties[]} */
+	const actions = [];
```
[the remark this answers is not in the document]
beejjacobs: I've removed `manage` and `add` in the latest change, as it was the weird `mapActionComponent` logic that I realised wasn't needed. Instead we pass in `userHasEditPermission` and only add the actions if they do.

I think `actions` is clear here, and the type is clearly stated above. Generally I don't advise adding types to variable names, as then you end up with things like `myBool` which doesn't add anything as the type system tells you that.

beejjacobs:
> Looking great overall. In spirit I agree with the changes and it's nice to see duplication being reduced. My only major concern is around naming of functions and variables not being descriptive enough or specific enough. More verbose code is more cumbersome to work with, but it's also easier to understand for someone less familiar with the codebase. When doing this type of refactoring, we risk fixing one problem by creating another worse problem (for me personally, trying to decipher how something works is generally more difficult than wading through duplicated code). It's really important that we preserve as much clarity as possible, and in my view proper naming is critical to that. Apologies if this seems like a pedantic point; I just feel strongly about this topic!

Thanks James. I think a big issue with this type of repeated code is when there are differences you don't know if its deliberate. We've already spotted differences between the lpaq mapper and the appeal one; some consistent functions to use takes away the ambiguity and makes the code much clearer and more consistent.

appeals/web/src/server/lib/mappers/appellantCase.mapper.js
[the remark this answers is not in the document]
james-hill-kc: I think this would be a bit clearer and more consistent if we just passed those values as parameters, then it wouldn't be dependent on the scope

appeals/web/src/server/lib/mappers/lpaQuestionnaire.mapper.js
james-hill-kc: I guess clarity is subjective, personally I think this approach is sacrificing clarity for brevity, which is a trend I generally disagree with in principle - it's trivial in this case, but in my experience, a lot of small deviations from the established convention which individually seem trivial can add up to a significant drop in clarity overall. But it's not a hill I'm willing to die on in this case

appeals/web/src/server/lib/page-components/boolean.js
james-hill-kc: Have you seen Lucy's PR https://github.com/Planning-Inspectorate/appeals-back-office/pull/432? there are plenty of examples in there of instances where a summaryListItem helper would save quite a bit of code (this would also reduce errors in the same way as the `yesNoInput` helper you added recently). That would be one for a separate PR regardless.

appeals/web/src/server/lib/page-components/document.js
james-hill-kc: Agreed. I think far-reaching changes in convention like this should be discussed or at least communicated to the team so we're aware this is what we should be doing when writing new code. Also we should open a tech debt ticket to ensure existing code not using this approach is refactored, or we'll end up with inconsistency, which is arguably worse than one consistent non-ideal approach

beejjacobs: Yes there is always a trade off between going in a new direction and having differing approaches, or sticking with what is there. I think in this instance the two can co-exist without issue, because we type-check function calls. Given how much clearer it makes these functions, it makes sense to use named parameters here. We don't need to refactor the whole codebase for this case, and for many functions it won't matter. If there are some key functions that would benefit, as in that ticket, it may make sense to refactor them, but I would keep it targeted.

james-hill-kc:
> > Looking great overall. In spirit I agree with the changes and it's nice to see duplication being reduced. My only major concern is around naming of functions and variables not being descriptive enough or specific enough. More verbose code is more cumbersome to work with, but it's also easier to understand for someone less familiar with the codebase. When doing this type of refactoring, we risk fixing one problem by creating another worse problem (for me personally, trying to decipher how something works is generally more difficult than wading through duplicated code). It's really important that we preserve as much clarity as possible, and in my view proper naming is critical to that. Apologies if this seems like a pedantic point; I just feel strongly about this topic!
> 
> Thanks James. I think a big issue with this type of repeated code is when there are differences you don't know if its deliberate. We've already spotted differences between the lpaq mapper and the appeal one; some consistent functions to use takes away the ambiguity and makes the code much clearer and more consistent.

Yep absolutely agree, as long as the naming is clear enough I'm happy :)

james-hill-kc (approved):

PR #532, opened 2024-10-09
Feature/a2 588 copy

HadrianHughesKC:
## Describe your changes

* feat(api): return comment counts for each status in Appeal response
* test(api): update tests to expect commentCounts property in response
* chore(api): add a TODO comment for a redundant if-statement
* refactor(web): use ternary operator to check isCaseOfficer
Just a minor improvement to readability by reducing nesting.
* feat(web): implement logic for Action required tab for statements

### Explanation for the `groupBy` query

The ACs for this ticket require us to check whether each appeal has any comments in `awaiting_review` state. I spent some time trying to figure out the best way to do this. We're fetching a list of stubs for the appeals which could be quite long, so we ideally want to avoid having to make a separate API call for every item. It would be better if the API just returned the information as part of the request for the appeals list.

On the other hand, including every representation for every appeal in the list could potentially add a huge amount of redundant data to the response. I decided the best way would be to do an aggregate query to just count the number of comments in each state. Then in the `web` code we can just check if there are more than 0 `awaiting_review` representations.

I didn't want to abstract this code too much before we know if it's necessary, but I've tried not to make it too specific. E.g. you can pass the representation type to the new repository method so it works for more than just comments.

[A2-588](https://pins-ds.atlassian.net/browse/A2-588)

## Type of change 🧩

- [ ] Documentation update

[A2-588]: https://pins-ds.atlassian.net/browse/A2-588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/setup-tests.js
+++ b/appeals/api/setup-tests.js
@@ -6,6 +6,7 @@ import { NODE_ENV_PRODUCTION } from '#endpoints/constants.js';
 const mockValidateBlob = jest.fn().mockResolvedValue(true);
 const mockRepGetById = jest.fn().mockResolvedValue({});
 const mockRepUpdateById = jest.fn().mockResolvedValue({});
+const mockRepGroupBy = jest.fn().mockResolvedValue([]);
 const mockAppealRelationshipAdd = jest.fn().mockResolvedValue({});
 const mockAppealRelationshipRemove = jest.fn().mockResolvedValue({});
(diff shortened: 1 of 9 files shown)


Si3rr4wow (approved):

BiggsKC (approved):
All good

PR #536, opened 2024-10-10
feature(web): create script to split large files

Si3rr4wow:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/scripts/src/split-file/cursor.js
+++ b/packages/scripts/src/split-file/cursor.js
@@ -1,0 +1,234 @@
+import { readFileSync } from 'fs';
+
+/**
+ * @typedef {'none' | 'in-import' | 'in-type' | 'in-codeblock' | 'end'} CursorState
+ * @typedef {'empty' | 'filling' | 'contains-import' | 'contains-type' | 'contains-codeblock'} FragmentState
+ */
+
+const IMPORT_START_INDICATOR = Buffer.from('import');
+const TYPE_START_INDICATOR = Buffer.from('/**');
+const BLOCK_START_INDICATORS = ['const', 'export', 'async', 'function'].map((str) =>
+	Buffer.from(str)
+);
+
+const IMPORT_END_INDICATOR = Buffer.from(';');
+const TYPE_END_INDICATOR = Buffer.from('*/');
+
+export class Cursor {
+	/** @type {CursorState} */
+	#state = 'none';
+
+	/** @type {number} */
+	#position = 0;
+
+	/** @type {Buffer} */
+	#fileBuffer;
+
+	/** @type {Buffer} */
+	#fragmentBuffer = Buffer.from('');
+
+	/** @type {FragmentState} */
+	#fragmentState = 'empty';
+
+	/** @param {string} filePath */
+	constructor(filePath) {
+		this.#fileBuffer = readFileSync(filePath);
+	}
+
+	logInternals() {
+		console.log('state: ', this.#state);
+		console.log('fragment state: ', this.#fragmentState);
+		console.log('position: ', this.#position);
+		console.log('fragmentBuffer: ', this.#fragmentBuffer.toString());
+	}
+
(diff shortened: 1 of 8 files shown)


packages/scripts/src/split-file/cursor.js
```diff
@@ -0,0 +1,234 @@
...
+const IMPORT_START_INDICATOR = Buffer.from('import');
+const TYPE_START_INDICATOR = Buffer.from('/**');
+const BLOCK_START_INDICATORS = ['const', 'export', 'async', 'function'].map((str) =>
+	Buffer.from(str)
+);
+
+const IMPORT_END_INDICATOR = Buffer.from(';');
+const TYPE_END_INDICATOR = Buffer.from('*/');
+
+export class Cursor {
```
Si3rr4wow, line 17: This guy goes through a file character by character and feeds back what kind of code it's reading (import, type declaration or code block). It fills an internal buffer with the element it's currently reading hen flags when that buffer contains something worth reading out. 

Currently it contains all the rules for distinguishing between different types of block, which makes it look quite gungey. I'd like to extract those but it didn't feel necessary now.

packages/scripts/src/split-file/get-code-sections.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+const writeTypesToBlocks = (blocks, typeStrings) =>
+	blocks.map(({ block, typeIndex }) => ({
+		block,
+		type: typeIndex !== null ? typeStrings[typeIndex] : null
+	}));
+/**
+ * @param {import("./cursor.js").Cursor} cursor
+ */
+export const getCodeSections = (cursor) => {
```
Si3rr4wow, line 29: Puts the various types of blocks into separate arrays for the file writer to use later

packages/scripts/src/split-file/get-paths.js
```diff
@@ -0,0 +1,24 @@
+import path from 'path';
+
+export const getPaths = () => {
```
Si3rr4wow, line 3: Reads the command line args and uses them to create useful paths

packages/scripts/src/split-file/write-files.js
```diff
@@ -0,0 +1,91 @@
+import { existsSync, mkdirSync, writeFileSync } from 'fs';
+import path from 'path';
+import { camelToKebab } from './utils.js';
+
+/**
+ * @param {string} str
+ * @returns {string}
+ */
+const extendPaths = (str) =>
```
Si3rr4wow, line 9: adds an extra `../` to paths so they point in the right direction

Si3rr4wow: Actually I realise now this is dependent on the output dir being one nest deep in the original files dir. I'll clamp down on output dir specification so this always works

packages/scripts/src/split-file/write-files.js
```diff
@@ -0,0 +1,91 @@
...
+	)}\n\n${blockWithTypesAndExport}`;
+
+	return dataToWrite;
+};
+
+/**
+ * @param {string} block
+ * @returns {string}
+ */
+const createBlockName = (block) => {
```
Si3rr4wow, line 34: Plucks the block's name from its declaration

fabfab-kc:
Hey @Si3rr4wow I don't think this is the place for it...
Unless I missed something, it's a local utility that may be useful to refactor code, but why in the `npm` scripts? I would have put it in a private repo / gists to make available to everyone, when needed?

Si3rr4wow:
@fabfab-kc I don't think, in its current state, this splitter has much value outside of this repo. As I mentioned in my comment on the `Cursor` this bakes in rules that are sympathetic to the way code is written for BO ie ES6 import syntax and jsdoc types. As it is this wouldn't even work on front office code. I can make it more configurable (as I said, I'd like to do that) but then this repo will still end up owning the config

PR #553, opened 2024-10-16
Feature/a2 1299 attachments

HadrianHughesKC:
## Describe your changes

This introduces a new API endpoint for creating IP comments. I've built this in a way that will be easy to extend to other types of representation (statements and final comments).

* fix(web): correct copy on add-ip-comment forms
* feat(packages/appeals): add a default folder for rep attachments
test(api): update tests to expect new default folder
* feat(api): create a route for creating representations
* feat(api): create a repository method to create representations
* feat(api): create a repository method to create ServiceUsers
* feat(api): refactor updateRepresentationById to update any field
* feat(api): create Representation and represented ServiceUser
* feat(api): create a repository method for adding attachments to reps
* feat(api): corrections to request body for create rep API
* feat(api): update body validator to match new input structure
* test(api): add tests for validator for creating IP comments
* feat(api): associate newly added documents when creating ip comments
* refactor(api): create Address prior to creating ServiceUser

Will squash commits before merging.

[A2-1299](https://pins-ds.atlassian.net/browse/A2-1299)

## Type of change 🧩

- [ ] Documentation update

[A2-1299]: https://pins-ds.atlassian.net/browse/A2-1299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appeals/api/src/server/repositories/service-user.repository.js
```diff
@@ -1,8 +1,19 @@
...
+ * @typedef {Object} ServiceUserInput
+ * @property {string} [organisationName]
+ * @property {string} [salutation]
+ * @property {string} [firstName]
+ * @property {string} [middleName]
+ * @property {string} [lastName]
+ * @property {string} [email]
+ * @property {string} [website]
+ * @property {string} [phoneNumber]
+ * @property {number} [addressId]
```
beejjacobs, line 15: I'd expect there to be Prisma types that cover this?

perhaps: `import('#db-client').Prisma.ServiceUserCreateInput` or similar

appeals/api/src/server/endpoints/representations/representations.service.js
```diff
@@ -96,13 +99,55 @@ export const updateRepresentationStatus = async (id, status, notes, reviewer) =>
...
+ * @property {string[]} attachments
+ * @property {string} redactionStatus
+ *
+ * @param {number} appealId
+ * @param {CreateRepresentationInput} input
+ * @returns {Promise<import('@pins/appeals.api').Schema.Representation>}
+ * */
+export const createRepresentation = async (appealId, input) => {
+	const { ipDetails, ipAddress } = input;
+
```
fabfab-kc, line 119: Probably the 3 repository calls below could be wrapped in a single transaction - maybe early optimisation we don't need at this stage...

appeals/api/src/server/openapi.json
```diff
@@ -3876,62 +3935,6 @@
 					}
 				}
 			}
-		},
```
fabfab-kc, line 3879: Not sure why this has gone... Worth checking with @AlexMoulin

packages/appeals/constants/documents.js
```diff
@@ -34,7 +34,8 @@ export const FOLDERS = [
 	`${APPEAL_CASE_STAGE.INTERNAL}/${APPEAL_DOCUMENT_TYPE.CROSS_TEAM_CORRESPONDENCE}`,
 	`${APPEAL_CASE_STAGE.INTERNAL}/${APPEAL_DOCUMENT_TYPE.INSPECTOR_CORRESPONDENCE}`,
 	`${APPEAL_CASE_STAGE.INTERNAL}/${APPEAL_DOCUMENT_TYPE.UNCATEGORISED}`,
-	`${APPEAL_CASE_STAGE.APPEAL_DECISION}/${APPEAL_DOCUMENT_TYPE.CASE_DECISION_LETTER}`
+	`${APPEAL_CASE_STAGE.APPEAL_DECISION}/${APPEAL_DOCUMENT_TYPE.CASE_DECISION_LETTER}`,
+	'representation/representationAttachments'
```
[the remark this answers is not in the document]
fabfab-kc: yep, will bring those to Gareth as part of the updates around the Rep schemas...

appeals/api/src/server/repositories/service-user.repository.js
fabfab-kc: I think this was pre-existing - I have a stash where I've started to go through but little progress so far...

appeals/api/src/server/repositories/document.repository.js
```diff
@@ -27,6 +27,22 @@ export const getDocumentById = (guid) => {
...
 };
 
+/**
+ * @param {string[]} guids
+ * @returns {Promise<Document[]>}
+ * */
+export const getDocumentsByIds = (guids) =>
+	Promise.all(
+		guids.map(async (guid) => {
+			const doc = await getDocumentById(guid);
```
fabfab-kc, line 37: Same here about performance. Probably this method could load them all by using something like
```
{ where: { guid in (guids) }}
```
Realistically there shouldn't be many attachments to a single comment, so again may be premature optimisation?

appeals/web/src/server/appeals/appeal-details/interested-party-comments/add-ip-comment/add-ip-comment.mapper.js
fabfab-kc, line 1: Generally, the upload settings are not yet in - no problem if the plan was to do them in another ticket, otherwise we could add now that we have the `representationAttachment` document type.

fabfab-kc (approved):
Aware of this being WIP and approving for the moment. However, have a look at the comments and maybe we can address something straight away...

appeals/api/src/server/repositories/document.repository.js
HadrianHughesKC: I considered this but I think doing it that way will mean if one of the GUIDs isn't found, it will fail silently, which we might not want.

appeals/api/src/server/endpoints/representations/representations.service.js
HadrianHughesKC: > Probably the 3 repository calls below could be wrapped in a single transaction - maybe early optimisation we don't need at this stage...

That's a good point, I did initially consider doing it all as a single Prisma query using nested `create` fields. The problem with both approaches is it introduces references to ServiceUser and Address into `representation.repository.js`, which we might want to keep solely focused on representations.

Might be worth the trade-off though, let me know what you think.

HadrianHughesKC: That's a good point, I did initially consider doing it all as a single Prisma query using nested create fields. The problem with both approaches is it introduces references to ServiceUser and Address into representation.repository.js, which we might want to keep solely focused on representations.

Might be worth the trade-off though, let me know what you think.

appeals/web/src/server/appeals/appeal-details/interested-party-comments/add-ip-comment/add-ip-comment.mapper.js
HadrianHughesKC: My thinking was to do it as a separate call to the API (set up in a separate PR) before making the call to create the Representation. I could create the documents as part of this API endpoint, but I want to make sure it's flexible so it can be used across different types of representation. So it just depends if there's ever going to be a scenario where we want to create a representation that links to already existing documents.

appeals/api/src/server/repositories/document.repository.js
fabfab-kc: good point!

packages/appeals/constants/documents.js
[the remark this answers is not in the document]
beejjacobs: OK, we need to note to revisit these then once there are constants for them.

appeals/api/src/server/endpoints/representations/representations.service.js
fabfab-kc: Probably we can go with this and then optimise later if we need. The nested fields may work better as they would be includes to the rep repository, as the rep schema still has dependencies on those, but happy to keep it as it is for now!

fabfab-kc (approved):
Let's go for it

PR #554, opened 2024-10-16
feature(web): add ip comments row to docs accordion

Si3rr4wow:
Ticket: https://pins-ds.atlassian.net/browse/A2-65

Adds IP comments row to S78 documents accordion on appeal detail page


appeals/web/src/server/lib/mappers/appeal/submappers/ip-comments.mapper.js
```diff
@@ -0,0 +1,27 @@
...
+					text: 'Interested party comments'
+				},
+				{
+					text: mapDocumentStatus(appealDetails?.documentationSummary?.ipComments?.status)
+				},
+				{
+					text: 'Not applicable'
+				},
+				{
+					html:
```
kandcdev, line 19: I think something like this would be a bit more readable:

```
const linkLabel = appealDetails?.documentationSummary?.ipComments?.status === 'received' ? 'Review' : 'Add';
...
html: `<a href="${currentRoute}/interested-party-comments" data-cy="review-ip-comments" class="govuk-link">${linkLabel} <span class="govuk-visually-hidden">I P comments</span></a>`;
```

Si3rr4wow: fair point, better now?

appeals/web/src/server/appeals/appeal-details/accordions/has.js
```diff
@@ -53,7 +52,24 @@ export function generateAccordion(appealDetails, mappedData, session, ipComments
 				}
 		  ];
 
-	const caseDocumentation = getCaseDocumentation(mappedData);
```
james-hill-kc, line 56: What's the reasoning behind removing `case-documentation.js` and moving the code inline here and in the matching s78 mapper?

appeals/web/src/server/lib/mappers/appeal/submappers/ip-comments.mapper.js
```diff
@@ -0,0 +1,28 @@
...
+				},
+				{
+					text: 'Not applicable'
+				},
+				{
+					html: `<a href="${currentRoute}/interested-party-comments" data-cy="review-ip-comments" class="govuk-link">${
+						appealDetails?.documentationSummary?.ipComments?.status === 'received'
+							? 'Review'
+							: 'Add'
+					} <span class="govuk-visually-hidden">I P comments</span></a>`
```
james-hill-kc, line 23: Very minor, the convention is for the visually hidden text to replicate the row label text ("Interested party comments" in this case)

james-hill-kc (commented):
Looks good, I've just left a couple of comments - one question and one very minor.

BiggsKC:
Could we please get a small PR description and link to the ticket?

appeals/web/src/server/appeals/appeal-details/accordions/has.js
Si3rr4wow: This is the first piece of a lot of divergence that's going to happen between HAS and S78, there's lots more documentation for the latter

kandcdev (approved):

PR #567, opened 2024-10-21
feat(web): add radio button to review page

Si3rr4wow:
[Ticket](https://pins-ds.atlassian.net/browse/A2-1109)

- Adds "Comment valid but requires redaction" radio button for IP comments from FO
- Separates view and review mappers into separate files with shared dependencies
- Adds higher order render function for view and review


appeals/web/src/server/appeals/appeal-details/interested-party-comments/view-and-review/view-and-review.controller.js
```diff
@@ -6,50 +6,45 @@ import {
...
  *
- * @param {import('@pins/express/types/express.js').Request} request
- * @param {import('@pins/express/types/express.js').RenderedResponse<any, any, Number>} response
+ * @param {(appealDetails: Appeal, comment: Representation) => PageContent} contentMapper
+ * @param {string} templatePath
+ * @returns {import('@pins/express').RenderHandler<unknown>}
  */
-export async function renderViewInterestedPartyComment(request, response) {
-	const { errors, currentComment } = request;
+const render = (contentMapper, templatePath) => (request, response) => {
```
Si3rr4wow, line 18: Feel free to poopoo this if it seems overcomplicated, it just seemed like an obvious factorisation to me

james-hill-kc: This is fine in principle, my only concern is we're not using this pattern anywhere else. It might just be that there's no use case for it anywhere else though. Either way, happy for this to stay in.

appeals/web/src/server/appeals/appeal-details/interested-party-comments/view-and-review/mapper/common.js
james-hill-kc, line 1: It makes sense to split out this shared code, but rather than having the `mapper` subdirectory (a pattern we're not using anywhere else as far as I'm aware), it would be more consistent with the existing patterns to put this either in `lib/mappers` or further up the feature directory tree in `interested-party-comments/interested-party-comments.mapper.js` (the second seems more preferable to me as it's closer to the other files)

james-hill-kc (commented):
Looks good overall, just one comment from me

appeals/web/src/server/appeals/appeal-details/interested-party-comments/view-and-review/view-and-review.controller.js
Si3rr4wow: I'm pretty sure this specific expression only exists here, because 404ing if there's no comment data doesn't really make sense anywhere else. I wouldn't really call this a patter, it's just writing `zx + yx` as `(z + y)x`

appeals/web/src/server/appeals/appeal-details/interested-party-comments/view-and-review/mapper/common.js
Si3rr4wow: I'll move them to lib, these were in the top level ip comment mapper until Friday when I moved them out because that file was getting large 😅

Si3rr4wow: Don't hate me, I've made this more complicated than it needs to be. I've put this in `page-components` under `#lib/mappers`. I also moved some appellant case mappers in there to help contextualise it. This is an increment towards fixing the thing we were talking about last week where we have two different kinds of mapper living in that directory. I'd also like to make an `instructions` dir for the other kind of mapper. I'll do more moving after this, just don't want to move everything all at once

james-hill-kc (approved):
LGTM!

PR #620, opened 2024-11-06
feat(appeals): investigate ability to bulk download

BenSurgisonPINS:
Bulk download spike

WEB:
- Add a link that downloads a zipped file containing all uploaded documents retaining the folder structure

[A2-1467 Spike - Investigate ability to bulk download from Azure storage](https://pins-ds.atlassian.net/browse/A2-1467)

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/app/components/file-downloader.component.js
+++ b/appeals/web/src/server/app/components/file-downloader.component.js
@@ -23,7 +23,7 @@ const validScanResult = APPEAL_VIRUS_CHECK_STATUS.SCANNED;
  * @param {SessionWithAuth} session
  * @returns {Promise<*>}
  */
-export const createBlobStorageClient = async (session) => {
+const createBlobStorageClient = async (session) => {
 	if (config.useBlobEmulator === true) {
 		return new BlobStorageClient(new BlobServiceClient(config.blobEmulatorSasUrl));
 	} else {
@@ -199,13 +199,14 @@ export const getDocumentDownloadByVersion = async ({ apiClient, params, session 
 };
 
 /**
+ * Create a blob download stream
  *
  * @param {import('@pins/blob-storage-client').BlobStorageClient} blobStorageClient
  * @param {string} blobStorageContainer
  * @param {string} blobStoragePath
  * @returns {Promise<*>}
  */
-export const createBlobDownloadStream = async (
+const createBlobDownloadStream = async (
 	blobStorageClient,
 	blobStorageContainer,
 	blobStoragePath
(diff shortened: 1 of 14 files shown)


appeals/web/src/server/app/components/file-downloader.component.js
```diff
@@ -198,3 +197,106 @@ export const getDocumentDownloadByVersion = async ({ apiClient, params, session
...
+					document.latestDocumentVersion;
+				return {
+					fullName: `${folder.path}/${document.name}`,
+					blobStorageContainer,
+					blobStoragePath,
+					documentURI
+				};
+			});
+		});
+
```
fabfab-kc, line 267: Here there should be a check for an empty `bulkFileInfo` (i.e. no files available).

appeals/web/src/server/app/components/file-downloader.component.js
```diff
@@ -198,3 +197,106 @@ export const getDocumentDownloadByVersion = async ({ apiClient, params, session
...
+	archive.pipe(response);
+
+	const blobStorageClient = await createBlobStorageClient(session);
+
+	if (!bulkFileInfo || !bulkFileInfo.length) {
+		return response.status(404);
+	}
+
+	const blobStreams = await Promise.all(
```
fabfab-kc, line 285: There's a small chance a document record may not have the corresponding blob (especially on dev machines): if that happens, the stream gets locked... maybe append to the archive only if the blob stream is valid?

appeals/web/src/server/app/components/file-downloader.component.js
```diff
@@ -198,3 +197,106 @@ export const getDocumentDownloadByVersion = async ({ apiClient, params, session
...
+ * Build a zipped file and Download
+ *
+ * @param {{apiClient: import('got').Got, params: {caseId: string, filename?: string}, session: SessionWithAuth}} request
+ * @param {Response} response
+ * @returns {Promise<Response>}
+ */
+export const getBulkDocumentDownload = async ({ apiClient, params, session }, response) => {
+	const { filename: requestedFilename, caseId } = params;
+
+	// console.log(JSON.stringify({ getBulkDocumentDownload1: { caseId, requestedFilename } }, null, 2));
```
fabfab-kc, line 249: remove

fabfab-kc (commented):
This looks great @BenSurgisonPINS - added a few comments on the best way to move it forward and take it out of draft

appeals/web/src/server/app/components/file-downloader.component.js
fabfab-kc: We should also have the ability to filter-out folders... e.g. internal and costs folders should probably be excluded (those will be the left part of the `folder.path` which usually indicates the area and the document type)

appeals/web/src/server/appeals/appeal-details/appeal-details.mapper.js
```diff
@@ -43,6 +52,7 @@ export async function appealDetailsPage(
 		...buildNotificationBanners(session, 'appealDetails', appealDetails.appealId),
 		...(await generateStatusTags(mappedData, appealDetails, request)),
 		generateCaseSummary(mappedData),
+		allDocumentsDownloadLink,
```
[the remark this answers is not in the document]
fabfab-kc: We would also to consider generally permissions, i.e. who can download the docs, or maybe even more in detail, i.e. which folder depending on permissions...

appeals/web/src/server/app/components/file-downloader.component.js
BenSurgisonPINS: I'm filtering out the empty blobStreams when appending to the archive to fix this issue.

BenSurgisonPINS: If there are any missing files, they will be reported within the missing-files.json file at the root of the zipped file

PR #629, opened 2024-11-08
refactor(web): create abstract mapping looper

Si3rr4wow:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
+++ b/appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
@@ -1,0 +1,26 @@
+/** @type {import('./types.d.ts').InitialiseAndMapDataFactory} */
+export const initialiseAndMapDataFactory = (
+	getSubmapperParams,
+	submaps,
+	getAppealType,
+	targetKey
+) => {
+	return (innerParams) => {
+		/** @type {MappedInstructions} */
+		const instructions = {};
+
+		const submappers = submaps[getAppealType(innerParams)];
+
+		const submapperParams = getSubmapperParams(innerParams);
+
+		Object.entries(submappers).forEach(([key, submapper]) => {
+			instructions[key] = submapper(submapperParams);
+		});
+
+		/** @type {Record<typeof targetKey, MappedInstructions>} */
+		// @ts-ignore
+		const mappedData = { [targetKey]: instructions };
+
+		return mappedData;
+	};
+};
--- a/appeals/web/src/server/lib/mappers/data/types.d.ts
+++ b/appeals/web/src/server/lib/mappers/data/types.d.ts
@@ -1,0 +1,13 @@
+import { MappedInstructions } from '../types';
+
+export type InitialiseAndMapDataFactory = <
+	TInnerParams extends Record<string, unknown>,
+	TSubMapperParams,
+	TSubmapper extends (params: TSubMapperParams) => MappedInstructions,
+	TTargetKey extends string
+>(
+	getSubmapperParams: (innerParams: TInnerParams) => TSubMapperParams,
+	submaps: Record<string, TSubMapper>,
+	getAppealType: (innerParams: TInnerParams) => string,
+	targetKey: TTargetKey
+) => (innerParams: TInnerParams) => Record<TTargetKey, MappedInstructions>;
--- a/appeals/web/src/server/appeals/appeal-details/lpa-questionnaire/lpa-questionnaire.mapper.js
+++ b/appeals/web/src/server/appeals/appeal-details/lpa-questionnaire/lpa-questionnaire.mapper.js
@@ -46,12 +46,12 @@ import { permissionNames } from '#environment/permissions.js';
  * @returns {Promise<PageContent>}
  */
 export async function lpaQuestionnairePage(lpaqDetails, appealDetails, currentRoute, session) {
-	const mappedLpaqDetails = initialiseAndMapLPAQData(
-		lpaqDetails,
+	const mappedLpaqDetails = initialiseAndMapLPAQData({
+		lpaQuestionnaireData: lpaqDetails,
 		appealDetails,
 		session,
 		currentRoute
-	);
+	});
 	const mappedAppealDetails = await initialiseAndMapAppealData(
 		appealDetails,
 		currentRoute,
--- a/appeals/web/src/server/appeals/change-page/change-page.mapper.js
+++ b/appeals/web/src/server/appeals/change-page/change-page.mapper.js
@@ -37,7 +37,12 @@ export async function appealChangePage(question, appealData, currentRoute, sessi
  * @returns {PageContent}
  */
 export function lpaQuestionnaireChangePage(question, appealData, lpaqData, session, currentRoute) {
-	const mappedData = initialiseAndMapLPAQData(lpaqData, appealData, session, currentRoute);
+	const mappedData = initialiseAndMapLPAQData({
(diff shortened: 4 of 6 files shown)


fabfab-kc:
Looks good to me - let's see if there's more comments incoming. 
The approach may seem hard to understand or maintain but the patterns are quite clear and it performs well!

appeals/web/src/server/lib/mappers/data/appeal/mapper.js
```diff
@@ -78,25 +82,35 @@ export async function initialiseAndMapAppealData(
...
+	if (!appealDetails.appealType) {
+		throw new Error('No appealType on appealDetails');
+	}
+	return appealDetails.appealType;
+};
 
-	return mappedData;
-}
+export const initialiseAndMapAppealData = initialiseAndMapDataFactory(
+	getSubmapperParams,
```
james-hill-kc, line 112: Would it be simpler to just pass the submapper params themselves rather than a function to generate them?

appeals/web/src/server/lib/mappers/data/appellant-case/mapper.js
```diff
@@ -23,44 +30,49 @@ const submaps = {
 };
 
 /**
- * @param {import('@pins/appeals.api').Appeals.SingleAppellantCaseResponse} appellantCaseData
- * @param {import('#appeals/appeal-details/appeal-details.types.js').WebAppeal} appealDetails
- * @param {import('../../../../app/auth/auth-session.service.js').SessionWithAuth} session
- * @param {string} currentRoute
- * @returns {MappedInstructions}
+ * @typedef {Object} InnerParams
```
james-hill-kc, line 33: This name is a bit vague, could it be updated to be more descriptive?

appeals/web/src/server/lib/mappers/data/appeal/mapper.js
```diff
@@ -78,25 +82,35 @@ export async function initialiseAndMapAppealData(
...
+	submaps,
+	getAppealType,
```
james-hill-kc, line 114: As above, it seems a little overcomplicated to pass a function here when I think you could just pass the params themseves

appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
```diff
@@ -0,0 +1,28 @@
+/** @type {import('./types.js').InitialiseAndMapDataFactory} */
+export const initialiseAndMapDataFactory = (
```
james-hill-kc, line 2: This name is a bit ambiguous especially if you're not very familiar with the codebase. I'd suggest maybe `getInitialiseAndMapDataFunction` or something similar

appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
```diff
@@ -0,0 +1,28 @@
...
+	getSubmapperParams,
+	submaps,
+	getAppealType,
+	targetKey
+) => {
+	return async (innerParams, filterKeys) => {
```
james-hill-kc, line 8: This is really nice in terms of the end result, but I worry about adding more layers of abstraction to our already multi-layered mapping system. I'm familiar with the codebase but I'm still struggling to wrap my head around all the moving parts in this function - do you think there's a way we could achieve the same result with simpler code that's easier to understand?

appeals/web/src/server/lib/mappers/data/types.d.ts
```diff
@@ -0,0 +1,16 @@
+import { MappedInstructions } from '../types';
+
+export type InitialiseAndMapDataFactory = <
+	TInnerParams extends Record<string, unknown>,
```
james-hill-kc, line 4: What's the reason for the `T` prefix here? I'd have thought this could just refer to the `InnerParams` type defined in TSDoc further up

appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
```diff
@@ -0,0 +1,28 @@
...
+		const submapperParams = await getSubmapperParams(innerParams);
+
+		Object.entries(submappers).forEach(([key, submapper]) => {
+			if (!filterKeys || filterKeys?.includes(key)) {
+				instructions[key] = submapper(submapperParams);
+			}
+		});
+
+		/** @type {Record<typeof targetKey, MappedInstructions>} */
+		// @ts-ignore
```
james-hill-kc, line 23: Is this `ts-ignore` still required?

appeals/web/src/server/lib/mappers/data/types.d.ts
```diff
@@ -0,0 +1,16 @@
+import { MappedInstructions } from '../types';
+
+export type InitialiseAndMapDataFactory = <
```
james-hill-kc, line 3: Similarly to my comment on `initialiseAndMapDataFactory`, the complexity of this type could be a worry in terms of understandability. It's true we're using TS already but there aren't many instances where it's this complex, and it's potentially a stumbling block for devs not as familiar with TS.

appeals/web/src/server/lib/mappers/data/appeal/mapper.js
james-hill-kc: That would also remove the need for `initialiseAndMapData` to be changed to async, I think

james-hill-kc (commented):
Nice work! The idea and end result is great, but I have some concerns about the implementation being a little on the complex side, details in my comments

appeals/web/src/server/lib/mappers/data/appeal/mapper.js
Si3rr4wow: I don't think this would simplify it. If we passes in params instead here's how I see that evolving: 
1. non-trivial param calculations for appeals mapper are copy and pasted into the 4 files where the mapper is used
2. this duplicate code looks messy, we create a function to create appeal submapper params
3. this function gets called in four files
The way I've done it here that function we'd end up creating is only needed in one place rather than in every place the mapper is used

appeals/web/src/server/lib/mappers/data/appeal/mapper.js
Si3rr4wow: We haven't worked out the appeal type when we call the mappers so that's not really possible. Also bear in mind with these outer params you don't need to supply then when you actually use the mapper, they're only important inside this file. We'll only need to access this file to add new appeal types, and that's just adding a new line to the submapper dictionary

appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
Si3rr4wow: I don't believe you could do this without a higher order function

appeals/web/src/server/lib/mappers/data/types.d.ts
Si3rr4wow: It's a generic, just a convention

appeals/web/src/server/lib/mappers/data/initialise-and-map-data.js
Si3rr4wow: Sadly yeah, it's not happy about this target key thing

appeals/web/src/server/lib/mappers/data/types.d.ts
Si3rr4wow: Gunna have to refer to Biggs' comment 
> more complex for noobs, but hey ho.

PR #709, opened 2024-11-29
API changes for statements and final comments

HadrianHughesKC:
## Describe your changes

* chore(packages/appeals): split FINAL_COMMENT type into 2 subtypes
There will be LPA final comments and appellant final comments.
* chore(api): change FINAL_COMMENT type to correct subtype in seed script
* feat(packages/appeals): add INCOMPLETE representation type
* feat(api): only allow LPA statements to be set to the incomplete status
* feat(api): create a single endpoint for fetching representations
* chore(api): generate swagger spec for new representations endpoint
* feat(api): add proper error handling for representations endpoint
* chore(api): remove representationType-specific endpoints
Representations can now be filtered by type from a single endpoint
* feat(api): enable passing multiple representation types
* refactor(web): change calls to representations API to match new contract

**Does include a breaking change to the existing representations API.**

[A2-1772](https://pins-ds.atlassian.net/browse/A2-1772)

## Type of change 🧩

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

[A2-1772]: https://pins-ds.atlassian.net/browse/A2-1772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HadrianHughesKC:
> It's a bit of rework I know but how would you feel about having one get endpoint for all reps that optionally takes a query param for type? It already accepts pagination vars and status so it's not that big of a step and would increase code reuse

I was initially thinking the same, but I noticed that for [this endpoint](https://github.com/Planning-Inspectorate/appeals-back-office/blob/8e2ec4cdc20662ed7183e6b8a3e9b71200217f90/appeals/api/src/server/endpoints/representations/representations.controller.js#L39) the behaviour and response is quite different. I'm just a bit concerned that these new endpoints will also end up sufficiently different from each other that a single endpoint would end up overloaded with parameters, and/or we'd end up splitting it back out into multiple endpoints.

What do you think?

Si3rr4wow:
Nice! I should have said I did a lil PR for this yesterday in case we did want to pivot to one endpoint but you've done more clean up so let's goo with this PR. Only thing I'd ask for is I'd set up the repository to accept multiple types, could we get that behaviour in here? Hopefully you can pilfer it https://github.com/Planning-Inspectorate/appeals-back-office/pull/722/files. 

Also I wasn't sure what the behaviour should be if no type was given. I went with giving all reps but maybe it should be none?

HadrianHughesKC:
> https://github.com/Planning-Inspectorate/appeals-back-office/pull/722/files

👍 That's set up now.

I think returning all types if no types are specified is the best approach. So the query string allows you to query an otherwise unfiltered set of results.

Si3rr4wow (approved):

PR #750, opened 2024-12-11
fix(api): importing duplicate documents

fabfab-kc:
Fixes import of duplicate documents, by appending a `_n` suffix in the file name.

Example input:
`
[
	{ documentType: 'A', originalFilename: 'name.pdf' },
	{ documentType: 'A', originalFilename: 'name.pdf' },
	{ documentType: 'A', originalFilename: 'name.pdf' }
]
`

Expected output:
`
[
	{ documentType: 'A', originalFilename: 'name.pdf' },
	{ documentType: 'A', originalFilename: 'name_1.pdf' },
	{ documentType: 'A', originalFilename: 'name_2.pdf' }
]
`

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/src/server/endpoints/integrations/integrations.utils.js
+++ b/appeals/api/src/server/endpoints/integrations/integrations.utils.js
@@ -1,0 +1,45 @@
+import Path from 'node:path';
+
+/** @typedef {import('pins-data-model').Schemas.AppellantSubmissionCommand['documents'][number]} AppellantSubmissionDocument */
+/** @typedef {import('pins-data-model').Schemas.LPAQuestionnaireCommand['documents'][number]} LPAQuestionnaireCommandDocument */
+
+/**
+ *
+ * @param {(AppellantSubmissionDocument|LPAQuestionnaireCommandDocument)[]} documents
+ * @returns {(AppellantSubmissionDocument|LPAQuestionnaireCommandDocument)[]}
+ */
+export const renameDuplicateDocuments = (documents) => {
+	const seen = new Set();
+
+	return documents.map((document) => {
+		let key = `${document.documentType}_${document.originalFilename}`;
+
+		if (!seen.has(key)) {
+			seen.add(key);
+			return document;
+		}
+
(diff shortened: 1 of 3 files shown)


appeals/api/src/server/endpoints/integrations/integrations.utils.js
```diff
@@ -0,0 +1,38 @@
...
+		let key = `${document.documentType}_${document.originalFilename}`;
+
+		if (!seen.has(key)) {
+			seen.add(key);
+			return document;
+		}
+
+		const nameParts = document.originalFilename.split('.');
+		const extension = nameParts.pop() || '';
+		const baseName = nameParts.join('.');
```
beejjacobs, line 22: Can we use `node:path` package instead for this logic? I think this logic is fine, but its clearer with methods like `path.basename` and `path.extname`.

appeals/api/src/server/endpoints/integrations/integrations.utils.js
```diff
@@ -0,0 +1,38 @@
...
+		let counter = 1;
+		let newFilename;
+
+		do {
+			newFilename = `${baseName}_${counter}.${extension}`;
```
beejjacobs, line 27: do we add `copy_${counter}` for clarity?

beejjacobs (commented):
LGTM, just a few tweaks to consider.

appeals/api/src/server/endpoints/integrations/integrations.utils.js
fabfab-kc: Hey Ben, sorry what do you mean here? to add 'copy' to the name? I thought would be less disruptive to just append the suffix, rather than changing the name even more...

kandcdev (approved):

appeals/api/src/server/endpoints/integrations/integrations.utils.js
beejjacobs: Sorry yes I was wondering if we added "copy" to the name to make it clearer for users?

kandcdev (approved):

PR #782, opened 2024-12-23
feat(api): refactoring of appeal details

fabfab-kc:
A general refactoring of the API code, currently only used in the loading and rendering of the appeal detail data.

## Type of change 🧩

- [ ] Documentation update


appeals/api/src/server/endpoints/appeal-details/routes.js
```diff
@@ -0,0 +1,32 @@
...
+		}
+		#swagger.responses[200] = {
+			description: 'Gets a single appeal by ID',
+			schema: { $ref: '#/components/schemas/Appeal' }
+		}
+		#swagger.responses[400] = {}
+		#swagger.responses[404] = {}
+	 */
+	validateRequest,
+	loadAppeal,
```
beejjacobs, line 28: Careful, this is async middleware without the `asyncHandler` wrapper - so promise rejections won't be handled.

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -0,0 +1,257 @@
...
+/** @typedef {import('@pins/appeals.api').Api.Folder} Folder */
+/** @typedef {{ appeal: Appeal, appealTypes: AppealType[]|undefined, context: keyof contextEnum|undefined }} MappingRequest */
+
+/**
+ *
+ * @param {MappingRequest} mappingRequest
+ * @returns {AppealDTO|undefined}
+ */
+export const mapCase = ({ appeal, appealTypes = [], context = contextEnum.appealDetails }) => {
+	if (context && appeal?.id && appeal?.caseCreatedDate) {
```
james-hill-kc, line 17: Is the idea here to pass in the whole appeal data and map only the required parts? That matches the approach we're using in web. Just wondering if there might be an optimisation where we could query only the required info from the DB? Probably premature though.

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -0,0 +1,257 @@
+import { apiMappers } from './api/index.js';
+import { contextEnum } from './context-enum.js';
+import { APPEAL_CASE_STAGE, APPEAL_CASE_TYPE, APPEAL_DOCUMENT_TYPE } from 'pins-data-model';
+
+/** @typedef {import('@pins/appeals.api').Schema.Appeal} Appeal */
+/** @typedef {import('@pins/appeals.api').Schema.AppealType} AppealType */
+/** @typedef {import('@pins/appeals.api').Api.Appeal} AppealDTO */
+/** @typedef {import('@pins/appeals.api').Api.Folder} Folder */
+/** @typedef {{ appeal: Appeal, appealTypes: AppealType[]|undefined, context: keyof contextEnum|undefined }} MappingRequest */
```
james-hill-kc, line 9: It would be helpful to have some documentation explaining how the mapping code is intended to be used, and how types like this one relate to that. It's probably a bit much to document inline within these type definitions.

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -0,0 +1,257 @@
...
+	if (context && appeal?.id && appeal?.caseCreatedDate) {
+		const caseMap = createCaseMap({ appeal, appealTypes, context });
+		return createCaseLayout(caseMap, { appeal, appealTypes, context });
+	}
+};
+
+/**
+ *
+ * @param {MappingRequest} mappingRequest
+ * @returns {Map<string, *>}
```
james-hill-kc, line 26: These type definitions could be more specific (it would also be nice to specify return types for the other functions that currently don't have them)

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -0,0 +1,257 @@
...
+		}
+	});
+}
+
+/**
+ *
+ * @param {Map<string, *>} caseMap
+ * @param {MappingRequest} mappingRequest
+ */
+function createCaseLayout(caseMap, mappingRequest) {
```
james-hill-kc, line 88: it's a pedantic point, but it might be better not to include `case` in the name of this function, as it seems to be mapping other things besides the case itself, depending on the input - I'm not sure if there's a generic term we're using to describe these pages, but maybe something like `createDataLayout` or `createFormattedData`?

james-hill-kc (commented):
Some really nice improvements here from my point of view. It's especially nice having all the definitions laid out so clearly. I've left a few comments. When possible, it would be really helpful to have a conceptual run-through of how this is intended to be used, to make sure everyone is aligned.

appeals/api/src/server/mappers/mapper-factory.js
fabfab-kc: yes I was looking at DB optimisations and done some on the appeal detail - it could be improved further though my next priority is the appeal lists, as those queries need to be improved, especially as we're adding more and more search facets.

appeals/api/src/server/mappers/mapper-factory.js
fabfab-kc: definitely, one of the goals was to improve the typing, which we could cascade to web and document in confluence once we're happy with it

appeals/api/src/server/mappers/mapper-factory.js
fabfab-kc: yeah that's the only cheat so far - was looking for suggestions if I could avoid having repeated definitions, like some magic ducktape :-)

appeals/api/src/server/mappers/mapper-factory.js
fabfab-kc: The idea is to evolve this so you can call with a context and will return only the part needed, e.g. questionnaire. Also to add a specific context for integrations to the mapping code will work similarly, with its own mappers.

appeals/api/src/server/endpoints/appeal-details/controller.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * @param {Request} req
+ * @param {Response} res
+ * @returns {Promise<Response>}
+ */
+const getAppeal = async (req, res) => {
+	const { appeal } = req;
+
+	const dto = await service.loadAndFormatAppeal({ appeal });
```
Si3rr4wow, line 14: strong ditto

appeals/api/src/server/endpoints/appeal-details/controller.js
```diff
@@ -0,0 +1,23 @@
...
+
+	return res.send({
+		...dto
+	});
```
Si3rr4wow, line 18: why do this spread rather than just `return res.send(dto)`?

appeals/api/src/server/endpoints/appeals.routes.js
```diff
@@ -36,6 +36,8 @@ import { listedBuildingRoutes } from './listed-buildings/listed-buildings.routes
 import { caseNotesRoutes } from './case-notes/case-notes.routes.js';
 import { representationRejectionReasonsRoutes } from './representation-rejection-reasons/representation-rejection-reasons.routes.js';
 
+import { default as appealDetailsRoutes } from './appeal-details/routes.js';
```
Si3rr4wow, line 39: Does `import appealDetailsRoutes from './appeal-details/routes.js';` not work here?

appeals/api/src/server/endpoints/appeals/appeals.controller.js
```diff
@@ -137,6 +137,7 @@ const getMyAppeals = async (req, res) => {
  */
 const getAppeal = async (req, res) => {
 	const { appeal } = req;
+	console.time('mapAppealByIdV1');
```
Si3rr4wow, line 140: Love that you perf tested this but do we want to keep this log?

appeals/api/src/server/mappers/api/s78/map-appellant-case.js
```diff
@@ -0,0 +1,16 @@
...
+/** @typedef {import('@pins/appeals.api').Api.AppellantCase} AppellantCase */
+/** @typedef {import('#mappers/mapper-factory.js').MappingRequest} MappingRequest */
+/**
+ *
+ * @param {MappingRequest} data
+ * @returns {AppellantCase|undefined}
+ */
+export const mapAppellantCase = (data) => {
+	const { appeal } = data;
+	const { appellantCase } = appeal;
```
Si3rr4wow, line 11:
```suggestion
	const { appeal: { appellantCase } } = data;
```

appeals/api/src/server/mappers/mapper-factory.js
Si3rr4wow, line 9: Does `keyof` work in this context? I have a feeling I've tried this before and it winds up as just `string`

appeals/api/src/server/mappers/mapper-factory.js
Si3rr4wow: This could be a fun thing to investigate. 

I'd've thought that there's a caching layer in the machine the DB runs on that will hash a query and cache the result against that for a short time. If that is the case then repeated queries for the entire appeal would be cheaper than different queries for just the parts we need on each page. 

Should be pretty easy to experiment with this.

appeals/api/src/server/endpoints/appeal-details/controller.js
fabfab-kc: leftover probably... will change

appeals/api/src/server/endpoints/appeals.routes.js
fabfab-kc: I'm going to try and split lists vs appeal details - temporarily this is to make them work side-by-side, but eventually we should end up with:

- appeal details (handling one appeal)
- appeal list (handling personal and national list)

Currently we have all that code in the same feature

appeals/api/src/server/mappers/mapper-factory.js
Si3rr4wow: Yeah I had a feeling this line would be in here :D 

On the frontend we were lucky to already have the `Instructions` and `PageComponent` types written by Alex. I think they're just big unions of all the component types? All the frontend mappers like this return one of those types. 

Maybe you could just compose a big union but I think it'll end up pretty non-specific anyway since some of these mappers return `string` or `undefined` and others return compound types. I think even the union of `Allocation` and `AppealDecision` will be `any`.

appeals/api/src/server/endpoints/appeals/appeals.controller.js
fabfab-kc: nope!

appeals/api/src/server/mappers/mapper-factory.js
fabfab-kc: it does!

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -0,0 +1,257 @@
...
+ * @returns {Map<string, *>}
+ */
+function createCaseMap(mappingRequest) {
+	const { appeal } = mappingRequest;
+
+	const casedata = createMap(apiMappers.apiSharedMappers, mappingRequest);
+	switch (appeal.appealType?.key) {
+		case APPEAL_CASE_TYPE.W: {
+			const s78 = createMap(apiMappers.apiS78Mappers, mappingRequest);
+			updateMap(casedata, s78);
```
Si3rr4wow, line 35: Does this work? Maybe I read `updateMap` wrong but can it add new keys to `casedata`?

fabfab-kc: this is to merge common vs appeal-type specific...

- common first
- the update with the specific mappers

Si3rr4wow (approved):

PR #793, opened 2025-01-06
fix(web): make show more wait for inner text

Si3rr4wow:
Still not totally sure why this bug happens but like a lot of those sorts of bugs it can be fixed by waiting for a bit 🤷🏻‍♀️

Adds a recursive timeout to the show more script that waits for inner text to appear in the target before proceeding.

### DIFF
--- a/appeals/web/src/client/components/show-more/show-more.js
+++ b/appeals/web/src/client/components/show-more/show-more.js
@@ -123,8 +123,6 @@ const awaitInnerText = async (
 const initialiseTextMode = async (/** @type {ShowMoreComponentInstance} */ componentInstance) => {
 	await awaitInnerText(componentInstance);
 
-	console.log('Setting attrs');
-
 	componentInstance.elements.root.setAttribute(
 		ATTRIBUTES.fullText,
 		componentInstance.elements.root.innerText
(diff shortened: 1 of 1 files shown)


kandcdev (approved):

glpins (dismissed):

appeals/web/src/client/components/show-more/show-more.js
```diff
@@ -103,7 +103,26 @@ const bindEvents = (/** @type {ShowMoreComponentInstance} */ componentInstance)
 		});
 };
 
-const initialiseTextMode = (/** @type {ShowMoreComponentInstance} */ componentInstance) => {
+const awaitInnerText = async (
+	/** @type {ShowMoreComponentInstance} */ componentInstance,
+	depth = 0
+) => {
+	if (componentInstance.elements.root.innerText || depth > 100) {
```
james-hill-kc, line 110: Just a suggestion, the depth limit and timeout duration could be moved into named variables

james-hill-kc (approved):
I left one comment, but this looks good to me. Were you able to replicate the issue? I remember struggling with that part

james-hill-kc (approved):

PR #839, opened 2025-01-16
feat(appeals): adds service bus topic and subs for reps

fabfab-kc:
Adds representation topic and subscriptions to the service bus

[Ticket](https://pins-ds.atlassian.net/browse/A2-2133)


bryn-pins (approved):
Seems fine to me, but haven't looked at BO terraform much before

PR #889, opened 2025-01-30
Feature/a2 2052 share reps

HadrianHughesKC:
## Describe your changes

* refactor(api): use BackOfficeAppError instead of custom error type
* fix(api): fix type error in updateRepresentation controller
* feat(api): error if trying to publish reps via the usual patch
* feat(api): create endpoint for publishing representations
* feat(api): implement publishing LPA statements
* feat(api): implement publishing final comments
* feat(api): broadcast after publishing representations

[A2-2052](https://pins-ds.atlassian.net/browse/A2-2052)

[A2-2052]: https://pins-ds.atlassian.net/browse/A2-2052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/endpoints/representations/representations.controller.js
+++ b/appeals/api/src/server/endpoints/representations/representations.controller.js
@@ -10,7 +10,8 @@ import {
 	DEFAULT_PAGE_NUMBER,
 	DEFAULT_PAGE_SIZE,
 	ERROR_NOT_FOUND,
-	ERROR_REP_ONLY_STATEMENT_INCOMPLETE
+	ERROR_REP_ONLY_STATEMENT_INCOMPLETE,
+	ERROR_REP_PUBLISH_USING_ENDPOINT
 } from '#endpoints/constants.js';
 import { getPageCount } from '#utils/database-pagination.js';
 import { Prisma } from '#utils/db-client/index.js';
@@ -19,6 +20,7 @@ import { EventType } from '@pins/event-client';
 import { createAuditTrail } from '#endpoints/audit-trails/audit-trails.service.js';
 import { camelToScreamingSnake } from '#utils/string-utils.js';
 import stringTokenReplacement from '#utils/string-token-replacement.js';
+import BackOfficeAppError from '#utils/app-error.js';
 
 /** @typedef {import('express').Request} Request */
 /** @typedef {import('express').Response} Response */
@@ -151,6 +153,10 @@ export async function updateRepresentation(request, response) {
 		return response.status(400).send({ errors: { status: ERROR_REP_ONLY_STATEMENT_INCOMPLETE } });
 	}
 
+	if (status === APPEAL_REPRESENTATION_STATUS.PUBLISHED) {
+		return response.status(400).send({ errors: { status: ERROR_REP_PUBLISH_USING_ENDPOINT } });
+	}
+
 	await representationService.updateRepresentation(parseInt(repId), request.body);
 
 	const updatedRep = await representationService.getRepresentation(parseInt(repId));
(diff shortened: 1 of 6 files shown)


appeals/api/src/server/endpoints/representations/representations.controller.js
```diff
@@ -159,22 +153,32 @@ export async function updateRepresentation(request, response) {
 		return response.status(400).send({ errors: { status: ERROR_REP_ONLY_STATEMENT_INCOMPLETE } });
 	}
 
+	if (
+		status === APPEAL_REPRESENTATION_STATUS.PUBLISHED &&
+		rep.representationType in
+			[APPEAL_REPRESENTATION_TYPE.LPA_STATEMENT, APPEAL_REPRESENTATION_TYPE.FINAL_COMMENT]
+	) {
+		return response.status(400).send({ errors: { status: ERROR_REP_PUBLISH_USING_ENDPOINT } });
+	}
```
Si3rr4wow, line 162: Why can these specific reps only be published from a dedicated endpoint?

appeals/api/src/server/endpoints/representations/representations.controller.js
```diff
@@ -261,3 +265,29 @@ export const updateRepresentationAttachments = async (req, res) => {
...
+	const publishedIds = await (async () => {
+		switch (query.type) {
+			case 'lpa_statement':
+				return [await representationService.publishLpaStatement(appeal)];
+			case 'final_comment':
+				return await representationService.publishFinalComments(appeal);
+			default:
+				throw new BackOfficeAppError(`invalid representation type parameter: ${query.type}`, 400);
+		}
+	})();
```
Si3rr4wow, line 286: I'm wondering if it'd be cleaner to have the switch return the function to be called, or even just had a dictionary of rep types to publish functions? There's just quite a lot of schmoo in here between the self calling, all the async await etc. This could just be

```
const dict = {
  lpa_statement: representationService.publishLpaStatement,
  final_comment: representationService.publishFinalComments
}

const publish = dict[query.type];

if (!publish) throw new BackOfficeAppError(`invalid representation type parameter: ${query.type}`, 400);

await publish(appeal);
```

appeals/api/src/server/endpoints/representations/representations.controller.js
```diff
@@ -261,3 +265,29 @@ export const updateRepresentationAttachments = async (req, res) => {
...
+ * @param {Response} res
+ * @returns {Promise<Response>}
+ */
+export async function publish(req, res) {
+	const { appeal, query } = req;
+
+	const publishedIds = await (async () => {
+		switch (query.type) {
+			case 'lpa_statement':
+				return [await representationService.publishLpaStatement(appeal)];
```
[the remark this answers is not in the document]
Si3rr4wow: oh, cus it's singular but final comments are plural? If you use my suggestion below you could fix this with ensure array on the result of `publish`

appeals/api/src/server/endpoints/representations/representations.controller.js
HadrianHughesKC: We want the LPA statement and final comments to be published as their own steps (the final comments all at once), and then we might transition the state of the appeal on the back of this (that will be a subsequent PR). My thinking was that allowing reps to be published individually from the usual patch endpoint could mess up the intended flow.

As for other rep types like IP comments, they may be added to this list. We might actually just want to prevent publishing altogether from this endpoint.

appeals/api/src/server/endpoints/representations/representations.controller.js
HadrianHughesKC: I'd agree except that `publishLpaStatement` returns a single ID, whereas `publishFinalComments` returns an array of IDs (that's why line 280 lifts the result into an array). I want to end up with an array of published IDs so I can run the broadcaster on them, but it seems wrong for `publishLpaStatement` to return an array itself when we know it'll be updating 1 rep.

appeals/api/src/server/endpoints/representations/representations.controller.js
```diff
@@ -261,3 +265,29 @@ export const updateRepresentationAttachments = async (req, res) => {
...
+ * @param {Request} req
+ * @param {Response} res
+ * @returns {Promise<Response>}
+ */
+export async function publish(req, res) {
+	const { appeal, query } = req;
+
+	const publishedIds = await (async () => {
+		switch (query.type) {
+			case 'lpa_statement':
```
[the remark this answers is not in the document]
HadrianHughesKC: Surprisingly there isn't, that's definitely an improvement we can make.

appeals/api/src/server/endpoints/representations/representations.controller.js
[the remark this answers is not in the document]
HadrianHughesKC: I personally think the current implementation is cleaner than ensuring array at the end. It's a minor thing though so willing to go either way.

appeals/api/src/server/endpoints/representations/representations.controller.js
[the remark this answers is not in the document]
fabfab-kc: what about this one: https://github.com/Planning-Inspectorate/appeals-back-office/blob/5b8cbfbbcf207fab763c423e6e5de84aaa0b3b05/packages/appeals/constants/common.js#L35

This is what I'm using to map from internal to data-model rep types...

fabfab-kc (approved):
The logic looks sound to me. Probably makes sense to always work on arrays/reptype combo as I'm sure in the near future there will be additional flows which may add more exchanges of data with the FO.

appeals/api/src/server/endpoints/representations/representations.controller.js
Si3rr4wow: I'm in support of fully preventing it here if it simplifies flows. This code will get ugly quickly if we keep adding more rep types to the publish endpoint

fabfab-kc (approved):

fabfab-kc (dismissed):

fabfab-kc (approved):

PR #908, opened 2025-02-05
fix(api): a2-1025 put back has notify case start

Si3rr4wow:
## Describe your changes

- Puts back HAS case start notify that I thought we wanted to swap this one for

[Ticket](https://pins-ds.atlassian.net/browse/A2-1025)


infrastructure/environments/prod.tfvars
```diff
@@ -27,8 +27,10 @@ apps_config = {
       appeal_confirmed_id                                             = "1776217e-b40d-4d78-82d1-8d881dcec897"
       appeal_incomplete_id                                            = "4001ac42-0d2f-4520-b1c0-cae481dd18ba"
       appeal_invalid_id                                               = "fdce6d3b-f712-4d12-94c6-e8a994fdbdaf"
-      appeal_start_date_change_appellant_id                           = "4240655a-802f-4a68-84ad-03deb4a3004f"
-      appeal_start_date_change_lpa_id                                 = "5fc727ad-0f7c-44fd-8f9a-038545a30991"
+      appeal_start_date_change_has_appellant_id                       = "76fc3001-5393-48ef-ae1a-377089e6ec5e"
+      appeal_start_date_change_has_lpa_id                             = "e238c035-e087-4e3e-ba37-a3869478fff1"
+      appeal_start_date_change_s78_appellant_id                       = "37367d1d-5254-4caf-a7b2-e1b4335e3f04"
```
fabfab-kc, line 32: So these are temporary right? Templates not yet on the notify prod service?

fabfab-kc (approved):
Just a comment around the TRAINING/PROD values...

infrastructure/environments/prod.tfvars
Si3rr4wow: afaik they are all on prod already? I don't have prod notify access I just get the GUIDs from Oli. Why?

fabfab-kc: because some have the same values as TEST/DEV...

DEV/TEST uses the IDs from the test notify, TRAINING/PROD from the production one

PR #913, opened 2025-02-06
fix(infrastructure): updating module version to 1.38

Medini-Aithal:
## Describe your changes
With infrastructure module v1.38, 
- Uses new public_network_access setting to ensure staging slots have the same networking rules set as the prod slot
- Deny all public network access for Api's
- Allow only front door access for Web Apps
- All site_config changes were ignored due to a incorrect ignore_changes entry. This has been updated to only ignore  the docker image name/tag.

This ensures IP restrictions allowing only front_door and default action settings will apply.

[Ticket-slot-access](https://pins-ds.atlassian.net/browse/DEV-308)
[Ticket-ignore-changes](https://pins-ds.atlassian.net/browse/DEV-351)


fabfab-kc (approved):
LG!

PR #933, opened 2025-02-12
Feature/a2 1142 notify

HadrianHughesKC:
## Describe your changes

* chore(api): add final comments done templates to config
* feat(api): send emails via notify when final comments are published
* chore(infrastructure): add template IDs to the infrastructure

[A2-1142](https://pins-ds.atlassian.net.mcas.ms/browse/A2-1142)

[A2-1142]: https://pins-ds.atlassian.net/browse/A2-1142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/endpoints/representations/representations.service.js
+++ b/appeals/api/src/server/endpoints/representations/representations.service.js
@@ -4,8 +4,14 @@ import neighbouringSitesRepository from '#repositories/neighbouring-sites.reposi
 import representationRepository from '#repositories/representation.repository.js';
 import serviceUserRepository from '#repositories/service-user.repository.js';
 import BackOfficeAppError from '#utils/app-error.js';
+import logger from '#utils/logger.js';
 import transitionState from '#state/transition-state.js';
-import { VALIDATION_OUTCOME_COMPLETE } from '#endpoints/constants.js';
+import {
+	ERROR_FAILED_TO_SEND_NOTIFICATION_EMAIL,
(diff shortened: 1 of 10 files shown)


appeals/api/src/server/endpoints/representations/representations.service.js
```diff
@@ -258,5 +264,85 @@ export async function publishFinalComments(appeal, azureAdUserId) {
...
 	return result;
 }
+
+/**
+ * @param {Appeal} appeal
+ * @param {import('#endpoints/appeals.js').NotifyClient} notifyClient
+ * @param {import('#endpoints/appeals.js').NotifyTemplate} template
+ * @param {string} recipientEmail
+ * */
+async function notifyFinalComments(appeal, notifyClient, template, recipientEmail) {
```
Si3rr4wow, line 293: Could these functions maybe be called `notifyFinalCommentsPublished` or something similar? Just so it's clear when we expect this to fire

Si3rr4wow (approved):
LGTM!

Not for this PR but wondering if we should look at moving the status change notify stuff I did the other week into `representationRepository.updateRepresentations` so it gets called by both the update rep end point and the publish endpoints? I think we could reduce the logical overhead of this feature if we did that

PR #948, opened 2025-02-14
Refactor/state machines

HadrianHughesKC:
## Describe your changes

* refactor(api): replace FPA and HAS state machines with a unified one
* feat(api): skip STATEMENTS and FINAL_COMMENTS states if HAS case
* feat(api): add valid appeal types for each state in state machine
* feat(api): only include valid states for appeal type in listStates


fabfab-kc (approved):
LGTM!

PR #960, opened 2025-02-17
fix(web): wrap redacted final comment in show-more component on view and review screen

james-hill-kc:
## Describe your changes

Adds a wrapping "show more" component to the redacted comment text summary list row on the view/review final comments screen. Also includes a small fix to the existing view-and-review unit tests (POST test was inside describe for GET tests, moved this out into its own describe)

https://pins-ds.atlassian.net/browse/A2-2393


BenSurgisonPINS (approved):
LGTM

PR #964, opened 2025-02-19
fix(api): handling null procedure types, defaults to written

fabfab-kc:
Fixes null procedure types in state machine.
[Ticket]()


HadrianHughesKC (approved):
LGTM

PR #966, opened 2025-02-19
fix(api): more handling of null procedures

fabfab-kc:
Missed one more check for procedure type in state machine transitions.


topickering (approved):

fabfab-kc:
Duplicate

PR #979, opened 2025-02-21
Fix/a2 1795 late entry

HadrianHughesKC:
## Describe your changes

* refactor(api): tidy up code in controller for adding documents
* fix(api): correct logic for determining late entries
An appellant case entry is late if the appellant case has already been
approved or rejected.
An LPA questionnaire entry is late if the questionnaire has already been
completed.

[A2-1795](https://pins-ds.atlassian.net/browse/A2-1795)


appeals/api/src/server/endpoints/documents/documents.service.js
```diff
@@ -423,27 +424,29 @@ export const addDocumentAudit = async (guid, version, auditTrail, action) => {
...
+async function isLateEntry(stage, appealId) {
 	switch (stage) {
-		case APPEAL_CASE_STAGE.APPELLANT_CASE:
-			return (
-				status !== APPEAL_CASE_STATUS.ASSIGN_CASE_OFFICER &&
-				status !== APPEAL_CASE_STATUS.VALIDATION &&
-				status !== APPEAL_CASE_STATUS.READY_TO_START
-			);
+		case APPEAL_CASE_STAGE.APPELLANT_CASE: {
+			const appelllantCase = await appellantCaseRepository.getByAppealId(appealId);
```
fabfab-kc, line 434: should have all this data in the appeal details, but maybe missing something.
I was starting to think whether we should add some micro-caching to the appeal detail, as it should now load most stuff and it's called a few times on each request

fabfab-kc (changes requested):

fabfab-kc (approved):
Nice!

PR #991, opened 2025-02-25
fix(web): bugfix/A2-2377-final-comments-fix-width

kandcdev:
https://pins-ds.atlassian.net/browse/A2-2377


james-hill-kc (approved):
LGTM!

PR #1014, opened 2025-02-28
fix(web): a2-2139 filter deleted docs from IP supporting docs row

edpackard-work:
## Describe your changes

Filter deleted items so they are not displayed in the IP supporting documents row.

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2139)


fabfab-kc (approved):
LGTM!

PR #1021, opened 2025-03-03
fix(api): broadcast of received reps dates

fabfab-kc:
Broadcasts the dates of the following representation fields with the appeal data:

- appellantCommentsSubmittedDate
- lpaCommentsSubmittedDate
- appellantStatementSubmittedDate
- lpaStatementSubmittedDate


bryn-pins (approved):
seems reasonable

PR #1024, opened 2025-03-03
Fix/a2 2435 prevent storing invalid agent or appellant

edpackard-work:
## Describe your changes

* add logic to only use session stored data if rendering errors on agent/appellant change page
* fix bug whereby organisation name 'inline' error message not rendering

[Ticket](https://pins-ds.atlassian.net/browse/A2-2435)


BenSurgisonPINS (approved):
lgtm

PR #1027, opened 2025-03-04
feat(web): feature/A2-1426-allocation-specialisms-updates

kandcdev:
https://pins-ds.atlassian.net/browse/A2-1426


appeals/api/src/database/seed/data-static.js
```diff
@@ -291,10 +291,10 @@ export const lpaQuestionnaireIncompleteReasons = [
  */
 
 export const specialisms = [
+	{ name: 'General allocation' },
```
fabfab-kc, line 294: This won't work

fabfab-kc (changes requested):
The ordering of the 'General allocation' needs to be done in the front end

glpins (dismissed):

fabfab-kc (approved):
LGTM!

PR #1036, opened 2025-03-05
fix(web): logic error in mapRepresentationDocumentSummaryActionLink

HadrianHughesKC:
## Describe your changes

Logic error causing the 'View' link to appear even when there's no LPA statement present.

[A2-2430](https://pins-ds.atlassian.net/browse/A2-2430)

[A2-2430]: https://pins-ds.atlassian.net/browse/A2-2430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/web/src/server/lib/representation-utilities.js
+++ b/appeals/web/src/server/lib/representation-utilities.js
@@ -8,41 +8,48 @@ export function isRepresentationReviewRequired(representationStatus) {
 }
 
 /**
+ * @typedef {'appellant-final-comments'|'lpa-final-comments'|'lpa-statement'} RepresentationType
+ *
  * @param {string} currentRoute
  * @param {string|undefined} documentationStatus
  * @param {string|null|undefined} representationStatus
- * @param {'appellant'|'lpa'|'lpa-statement'} finalCommentsType
+ * @param {RepresentationType} representationType
  * @returns {string} action link html
  */
 export function mapRepresentationDocumentSummaryActionLink(
 	currentRoute,
 	documentationStatus,
 	representationStatus,
-	finalCommentsType
+	representationType
 ) {
-	if (documentationStatus === 'received' && representationStatus) {
-		const reviewRequired =
-			representationStatus === APPEAL_REPRESENTATION_STATUS.AWAITING_REVIEW ||
-			representationStatus === APPEAL_REPRESENTATION_STATUS.INCOMPLETE;
-
-		if (finalCommentsType === 'lpa-statement') {
-			return `<a href="${currentRoute}/lpa-statement" data-cy="${
-				reviewRequired ? 'review' : 'view'
-			}-lpa-statement" class="govuk-link">${
-				reviewRequired ? 'Review' : 'View'
-			}<span class="govuk-visually-hidden"> LPA statement</span></a>`;
-		}
-
-		return `<a href="${currentRoute}/final-comments/${finalCommentsType}" data-cy="${
-			reviewRequired ? 'review' : 'view'
-		}-${finalCommentsType}-final-comments" class="govuk-link">${
-			reviewRequired ? 'Review' : 'View'
-		} <span class="govuk-visually-hidden">${
-			finalCommentsType === 'lpa' ? 'L P A' : finalCommentsType
-		} final comments</span></a>`;
+	if (documentationStatus !== 'received') {
+		return '';
 	}
 
-	return '';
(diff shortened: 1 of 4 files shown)


appeals/web/src/server/lib/representation-utilities.js
```diff
@@ -20,29 +20,29 @@ export function mapRepresentationDocumentSummaryActionLink(
...
-			}-lpa-statement" class="govuk-link">${
-				reviewRequired ? 'Review' : 'View'
-			}<span class="govuk-visually-hidden"> LPA statement</span></a>`;
-		}
+	const reviewRequired =
+		representationStatus === APPEAL_REPRESENTATION_STATUS.AWAITING_REVIEW ||
+		representationStatus === APPEAL_REPRESENTATION_STATUS.INCOMPLETE;
 
-		return `<a href="${currentRoute}/final-comments/${finalCommentsType}" data-cy="${
+	if (finalCommentsType === 'lpa-statement') {
```
james-hill-kc, line 31: If this variable is being used for things other than the final comment type, we should change the name to something more generic like `representationSubType` (it looks like this was pre-existing)

appeals/web/src/server/lib/representation-utilities.js
```diff
@@ -20,29 +20,29 @@ export function mapRepresentationDocumentSummaryActionLink(
...
-		if (finalCommentsType === 'lpa-statement') {
-			return `<a href="${currentRoute}/lpa-statement" data-cy="${
-				reviewRequired ? 'review' : 'view'
-			}-lpa-statement" class="govuk-link">${
-				reviewRequired ? 'Review' : 'View'
-			}<span class="govuk-visually-hidden"> LPA statement</span></a>`;
-		}
+	const reviewRequired =
+		representationStatus === APPEAL_REPRESENTATION_STATUS.AWAITING_REVIEW ||
+		representationStatus === APPEAL_REPRESENTATION_STATUS.INCOMPLETE;
```
james-hill-kc, line 29: Again it looks like this was pre-existing, but seems wrong. Incomplete suggests the rep has already been reviewed (although LPA statement puts 'incomplete' in the `status` field in this case, not `representationStatus`). What are the circumstances where the rep status would be 'incomplete'?

james-hill-kc (commented):
I've left a couple of comments. As far as I can tell, `mapRepresentationDocumentSummaryActionLink` is only called from the appellant/LPA final comments mappers, not from LPA statement code, which is strange considering this fixes a bug in the LPA statements row - maybe there's a change in `lpa-statement.mapper` that hasn't been committed? I know this is pre-existing code, but it feels like it should be revisited; the logic in the function could be easier to follow - although that's probably one to address separately as a tech debt item.

appeals/web/src/server/lib/representation-utilities.js
HadrianHughesKC: I did think that yeah, wasn't sure if there was something I was missing about how it works.

appeals/web/src/server/lib/representation-utilities.js
HadrianHughesKC: I refactored documentationSummary so the `status` field is always either `received` or `not_received`, including for LPA statement, so the `representationStatus` field is the right place to look.

The case can be progressed from STATEMENTS -> FINAL_COMMENTS if the LPA statement is incomplete, but I'm not certain if the LPA statement can be re-reviewed once it's been marked is incomplete. I might need to check that with the BAs.

HadrianHughesKC:
Thanks @james-hill-kc 

I've confirmed with Fred that it should be 'Review' when the LPA statement is marked as incomplete. Also refactored to make it make more sense for LPA statements. It seemed like they were a bit tacked onto this function as an afterthought.

appeals/web/src/server/lib/representation-utilities.js
james-hill-kc: Ah that makes sense - it's a nice improvement to have all the documentationSummary entries work the same way. I guess entries without a representationStatus (eg. lpa questionnaire) will still have "incomplete" in the status field (overriding the received / not_received) but not much we can do about that without adding a "reviewStatus" or similar, but that's definitely out of scope of this PR

james-hill-kc (approved):
Looks good to me!

PR #1037, opened 2025-03-05
fix(api): fix appeal formatting types

Si3rr4wow:
## Describe your changes

Corrects appeal formatter types in the API

[Ticket](https://pins-ds.atlassian.net/browse/A2-1362)

### DIFF
--- a/appeals/api/src/server/endpoints/appeals/__tests__/appeals.test.js
+++ b/appeals/api/src/server/endpoints/appeals/__tests__/appeals.test.js
@@ -1,3 +1,4 @@
+// @ts-nocheck
 import { jest } from '@jest/globals';
 import { request } from '../../../app-test.js';
 import {
--- a/appeals/api/src/server/endpoints/appeals/appeals.formatter.js
+++ b/appeals/api/src/server/endpoints/appeals/appeals.formatter.js
@@ -233,7 +233,7 @@ export const mapAppealToDueDate = (appeal, appellantCaseStatus, appellantCaseDue
 			}
 			if (appeal.siteVisit) {
 				return addBusinessDays(
-					new Date(appeal.siteVisit.visitEndTime || appeal.siteVisit.visitDate || ''),
+					new Date(appeal.siteVisit.visitEndTime || appeal.siteVisit.visitDate || 0),
 					approxStageCompletion.STATE_TARGET_ISSUE_DETERMINATION_AFTER_SITE_VISIT
 				);
 			}
@@ -268,7 +268,7 @@ export const mapAppealToDueDate = (appeal, appellantCaseStatus, appellantCaseDue
 			});
 		}
 		case APPEAL_CASE_STATUS.AWAITING_EVENT: {
-			return new Date(appeal.siteVisit?.visitDate || '');
+			return new Date(appeal.siteVisit?.visitDate || 0);
 		}
 		default: {
 			return undefined;
--- a/appeals/api/src/server/endpoints/appeals/appeals.formatter.js
+++ b/appeals/api/src/server/endpoints/appeals/appeals.formatter.js
@@ -34,9 +34,12 @@ const approxStageCompletion = {
 /** @typedef {import('#endpoints/appeals').DocumentationSummary} DocumentationSummary */
 /** @typedef {import('#db-client').AppealStatus} AppealStatus */
 /** @typedef {import('@pins/appeals.api').Schema.Representation} Representation */
+/** @typedef {import('#repositories/appeal-lists.repository.js').DBAppeals} DBAppeals */
+/** @typedef {DBAppeals[0]} DBAppeal */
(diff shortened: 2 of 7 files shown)


appeals/api/src/server/endpoints/appeals/__tests__/appeals.test.js
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
Si3rr4wow, line 1: Are we cool with this in tests? It seems like a pretty common practice and personally I'm not that picky about ignoring TS in tests

appeals/api/src/server/endpoints/appeals/appeals.formatter.js
```diff
@@ -230,7 +233,7 @@ export const mapAppealToDueDate = (appeal, appellantCaseStatus, appellantCaseDue
 			}
 			if (appeal.siteVisit) {
 				return addBusinessDays(
-					new Date(appeal.siteVisit.visitEndTime || appeal.siteVisit.visitDate),
+					new Date(appeal.siteVisit.visitEndTime || appeal.siteVisit.visitDate || 0),
```
Si3rr4wow, line 236: TS isn't keen on passing `null` to a `Date` constructor. `0` preserves the behaviour in the case that both `visitEndTime` and `visitDate` were null

appeals/api/src/server/repositories/appeal-lists.repository.js
```diff
@@ -19,9 +19,8 @@ import { getEnabledAppealTypes } from '#utils/feature-flags-appeal-types.js';
  * @param {number} inspectorId
  * @param {number} caseOfficerId
  * @param {boolean} isGreenBelt
- * @returns {Promise<[number, Omit<Appeal, 'parentAppeals' | 'childAppeals'>[]]>}
```
Si3rr4wow, line 22: I've exported the inferred return types of these funcs so the types are generated by Prisma

appeals/api/src/server/repositories/appeal-lists.repository.js
```diff
@@ -82,34 +81,44 @@ const getAllAppeals = (
 		})
 	};
 
-	return databaseConnector.$transaction([
```
Si3rr4wow, line 85: I didn't see the point in this being a transaction?

appeals/api/src/server/endpoints/appeals/__tests__/appeals.test.js
HadrianHughesKC: Yeah no big deal imo

HadrianHughesKC (dismissed):
Left a few comments but it all looks good.

The only thing that stands out to me is the union types like `DBAppeal | DBUserAppeal`. It looks like in these cases it works fine but they can get annoying to work with if you want to select a property that's on one member of the union and not the other. You have to start introspecting so you can downcast to the specific type.

appeals/api/src/server/endpoints/appeals/appeals.formatter.js
Si3rr4wow: I mean 1/1/1970 is probably not the correct answer but I'm doing this PR to fix the types, not the code. The 0 preserves the behaviour and satisfies ts.

Screenshot 2025-03-06 at 09 30 35

appeals/api/src/server/repositories/appeal-lists.repository.js
```diff
@@ -82,34 +81,44 @@ const getAllAppeals = (
...
+
+	return appeals;
 };
 
 /**
  * @param {string} userId
  * @param {number} pageNumber
  * @param {number} pageSize
  * @param {string} [status]
- * @returns {Promise<[number, Omit<Appeal, 'parentAppeals' | 'childAppeals'>[], *[]]>}
```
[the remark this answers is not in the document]
Si3rr4wow: I think it's better to use the inferred type since this function returns a Prisma function with a complex include, so the return type is also very complex

appeals/api/src/server/endpoints/appeals/appeals.service.js
```diff
@@ -124,13 +130,16 @@ const mapInspectors = async (appeals) => {
 			return inspectorList;
 		}
 		return [...inspectorList, inspector];
-	}, []);
+	}, initialInspectors);
```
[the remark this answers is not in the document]
Si3rr4wow: I feel like I've tried that before and it's been unhappy with me but I'll give it a shot :)

Si3rr4wow:
@HadrianHughesKC yeah I hear you on the union types but these functions were pre-existing and already handled two different types, just unknowingly because their typings were wrong. If the formatting of the types diverges in future the dev doing that still has the option of duplicating the existing function and modifying one copy to suit the divergent need. 

I'm trying to avoid modifying the code as much as possible here since the task is just to solve the ts errors. Hope that makes sense :)

toyin-pins (approved):

PR #1040, opened 2025-03-06
Dynamic Back Button and Redirect URL Functionality

kandcdev:
https://pins-ds.atlassian.net/browse/A2-2401

This update introduces dynamic back button behavior and flexible redirect URL handling:
- Reads the `backUrl` query parameter and processes it using the `constructURL` utility
- Since most of the routes are sub-routes of appeal details page - instead of passing the full URL every time, only the path **after** `/<appealId>/` needs to be provided.  
  - **Example:** `?backUrl=/share` generates `/appeals-service/appeal-details/${appealId}/share`.  
- Routes that don’t follow this pattern, such as **Personal List**, are handled via a predefined mapping
  - **Example:** `?backUrl=/personal-list` resolves to `/appeals-service/personal-list`.  
- The same logic can be applied to **redirect URLs** after an action.  
  - **Example:** Passing `?redirectTo=/personal-list` redirects to **Personal List** instead of the default **Appeal Details** page.  

Thoughts or suggestions are welcome!

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/representations/__tests__/__snapshots__/representations.test.js.snap
+++ b/appeals/web/src/server/appeals/appeal-details/representations/__tests__/__snapshots__/representations.test.js.snap
@@ -9,7 +9,7 @@ exports[`representations GET /share should contain link to interested-party-comm
     </div>
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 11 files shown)


glpins (dismissed):

kandcdev:
> url-parser file is just empty, is that intentional or can it just be deleted?

good spot! removed empty file

toyin-pins (approved):

PR #1043, opened 2025-03-07
feat(api): pass development type in appellant case get and accept in patch

Si3rr4wow:
## Describe your changes

Adds get and patch capabilities to developmentType field on appellant cases

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2367)


fabfab-kc (approved):
LGTM!

PR #1044, opened 2025-03-07
feat(api): A2-2492 Add appeal type to list of filters in the all cases list

glpins:
## Describe your changes

Added appealTypeId as a filter to the GET appeals endpoint and updated tests to support it.

[A2-2492](https://pins-ds.atlassian.net/browse/A2-2492)

[A2-2492]: https://pins-ds.atlassian.net/browse/A2-2492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/repositories/appeal-lists.repository.js
+++ b/appeals/api/src/server/repositories/appeal-lists.repository.js
@@ -19,6 +19,7 @@ import { getEnabledAppealTypes } from '#utils/feature-flags-appeal-types.js';
  * @param {number} inspectorId
  * @param {number} caseOfficerId
  * @param {boolean} isGreenBelt
+ * @param {number} appealTypeId
  * @returns {Promise<[number, Omit<Appeal, 'parentAppeals' | 'childAppeals'>[]]>}
  */
 const getAllAppeals = (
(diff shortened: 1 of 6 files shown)


appeals/api/src/server/repositories/appeal-lists.repository.js
```diff
@@ -79,6 +81,9 @@ const getAllAppeals = (
 		}),
 		...(!!caseOfficerId && {
 			caseOfficerUserId: caseOfficerId
+		}),
+		...(!!appealTypeId && {
+			appealTypeId
 		})
```
Si3rr4wow, line 87: The appealTypeId of zero is not valid I assume?

Si3rr4wow (dismissed):

appeals/api/src/server/repositories/appeal-lists.repository.js
glpins: Since the frontend always provides valid values, the existing approach is fine. The implementation should remain consistent with how other filters are handled.

Si3rr4wow (approved):

PR #1045, opened 2025-03-07
fix(web): fixing read more text link to not show when there are less than 300 chars

toyin-pins:
## Describe your changes
- Returning early from initialising instance statement so that read more link text doesn't show when less than 300 chars.  

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2454)

### DIFF
--- a/appeals/web/src/client/components/show-more/show-more.js
+++ b/appeals/web/src/client/components/show-more/show-more.js
@@ -141,14 +141,14 @@ const initialiseTextMode = async (/** @type {ShowMoreComponentInstance} */ compo
 		componentInstance.elements.root.innerText
 	);
 
-	componentInstance.elements.root.innerHTML = '';
-	componentInstance.elements.root.appendChild(contentSpan);
-
 	const ellipsisSpan = document.createElement('span');
 
 	ellipsisSpan.className = CLASSES.ellipsis;
 	ellipsisSpan.setAttribute('aria-hidden', 'true');
 	ellipsisSpan.textContent = '…';
+
+	componentInstance.elements.root.innerHTML = '';
+	componentInstance.elements.root.appendChild(contentSpan);
 	componentInstance.elements.root.appendChild(ellipsisSpan);
 };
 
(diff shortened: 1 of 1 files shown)


appeals/web/src/client/components/show-more/show-more.js
```diff
@@ -190,26 +196,31 @@ const initialiseComponentInstance = async (
 		await initialiseTextMode(componentInstance);
 	}
 
-	const button = document.createElement('button');
-	button.className = CLASSES.toggleButton;
-	button.setAttribute('aria-expanded', 'false');
-	button.setAttribute('type', 'button');
+	if (
+		componentInstance.elements.root.getAttribute(ATTRIBUTES.fullText).length >
+		componentInstance.options.maximumCharactersBeforeHiding
```
james-hill-kc, line 201: This fix will definitely work, and there's nothing wrong with this approach. But a simpler fix might be to move this conditional check into `initialiseComponentInstance`, just below the call to `initialiseOptions(componentInstance);` - if the full text isn't longer than the maximumCharactersBeforeHiding, you could just return without initialising the component, which should have the same effect

james-hill-kc: For example: ```const initialiseComponentInstance = async (
	/** @type {ShowMoreComponentInstance} */ componentInstance
) => {
	initialiseOptions(componentInstance);

	if (componentInstance.elements.root.getAttribute(ATTRIBUTES.fullText).length <=
	componentInstance.options.maximumCharactersBeforeHiding) {
		return;
	}```

james-hill-kc: const initialiseComponentInstance = async (
	/** @type {ShowMoreComponentInstance} */ componentInstance
) => {
	initialiseOptions(componentInstance);

	if (componentInstance.elements.root.getAttribute(ATTRIBUTES.fullText).length <=
	componentInstance.options.maximumCharactersBeforeHiding) {
		return;
	}

james-hill-kc (commented):
Looks good, but I think there might be a simpler approach - details in my comments

appeals/web/src/client/components/show-more/show-more.js
toyin-pins: That is a much more elegant solution thank you!

james-hill-kc (approved):
Looks good to me!

PR #1058, opened 2025-03-11
fix(appeals) full text width on review lpa statements view (a2-2392)

pinsMT:
Change:
Created condition which instruct IP comments view to use two-thirds content size and inserts style value into html attributes.

Ticket:
https://pins-ds.atlassian.net/browse/A2-2392

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
+++ b/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
@@ -22,10 +22,17 @@ export const renderReviewLpaStatement = async (request, response) => {
 		backUrl
 	);
 
-	return response.status(200).render('patterns/display-page.pattern.njk', {
-		errors,
-		pageContent
-	});
+	return response
+		.status(200)
+		.render(
+			isReview
+				? 'patterns/change-page-full-width.pattern.njk'
+				: 'patterns/display-page.pattern.njk',
+			{
+				errors,
+				pageContent
+			}
+		);
 };
 
 /**
--- a/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
+++ b/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
@@ -22,12 +22,10 @@ export const renderReviewLpaStatement = async (request, response) => {
 		backUrl
 	);
 
-	return response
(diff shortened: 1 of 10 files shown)


appeals/web/src/server/views/patterns/change-page-full-width.pattern.njk
```diff
@@ -0,0 +1,33 @@
+{%- extends "app/layouts/app.layout.njk" -%}
```
james-hill-kc, line 1: This pattern template looks identical to `change-page.pattern.njk` apart from using a different layout. It would be good to avoid the duplication if possible. The duplicated parts could probably be moved into an `include` and imported into `change-page.pattern.njk` and `change-page-full-width.pattern.njk`.

Si3rr4wow (approved):

appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
```diff
@@ -24,10 +24,15 @@ export const renderReviewLpaStatement = async (request, response) => {
 
 	return response
 		.status(200)
-		.render(isReview ? 'patterns/change-page.pattern.njk' : 'patterns/display-page.pattern.njk', {
-			errors,
-			pageContent
-		});
+		.render(
+			isReview
+				? 'patterns/change-page-full-width.pattern.njk'
```
james-hill-kc, line 29: This could be simplified by only using `display-page.pattern.njk` regardless of whether the page is in 'view' or 'review' mode. Strictly speaking, the change page pattern is only intended to be used for change pages. For review pages (like the appellant case and LPA Questionnaire), the display page pattern is used. That would remove the need for the full width variant of the change page pattern template.

james-hill-kc (commented):
I've reviewed and left a couple of comments

appeals/web/src/server/views/patterns/change-page-full-width.pattern.njk
pinsMT, line 1: page htl content contained within file and imported with include

appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/lpa-statement.controller.js
pinsMT: Possibly create separate pattern template in the future? tech debt ticket?

james-hill-kc (approved):
LGTM!

PR #1063, opened 2025-03-12
fix(web): a2-2563 wrap long redacted strings in shared ip comments ta…

edpackard-work:
…ble cell

* for display purposes only, add invisible spaces after each redaction character to prevent long redaction strings from overspilling gov uk table cells

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2563)


james-hill-kc (approved):
Looks good to me!

PR #1068, opened 2025-03-12
Fix/a2 2556 display ip comment address from fo

edpackard-work:
[Ticket]()

### DIFF
--- a/appeals/api/src/server/endpoints/integrations/integrations.broadcasters/service-users.js
+++ b/appeals/api/src/server/endpoints/integrations/integrations.broadcasters/service-users.js
@@ -19,26 +19,15 @@ export const broadcastServiceUser = async (userId, updateType, roleName, caseRef
 	if (!config.serviceBusEnabled && config.NODE_ENV !== 'development') {
 		return false;
 	}
-	const user = await databaseConnector.serviceUser.findUnique({
-		where: { id: userId }
-	});
+
+	const user = await getServiceUser(userId);
 
 	if (!user) {
 		pino.error(`Trying to broadcast info for service-user ${userId}, but it was not found.`);
 		return false;
 	}
 
-	let userAddress;
-	if (user.addressId) {
-		userAddress = await databaseConnector.address.findUnique({
-			where: { id: user.addressId }
-		});
-		if (!userAddress) {
-			userAddress = undefined;
-		}
-	}
-
-	const msg = mapServiceUserEntity(user, userAddress, roleName, caseReference);
+	const msg = mapServiceUserEntity(user, roleName, caseReference);
 
 	if (msg) {
 		const validationResult = await validateFromSchema(schemas.events.serviceUser, msg);
@@ -64,3 +53,16 @@ export const broadcastServiceUser = async (userId, updateType, roleName, caseRef
 
 	return false;
 };
+
+/**
+ *
+ * @param {number} userId
+ */
+const getServiceUser = (userId) => {
+	return databaseConnector.serviceUser.findUnique({
+		where: { id: userId },
+		include: { address: true }
+	});
+};
+
+/** @typedef {Awaited<ReturnType<getServiceUser>>} GetServiceUser */
--- a/appeals/api/src/server/endpoints/integrations/integrations.broadcasters/service-users.js
+++ b/appeals/api/src/server/endpoints/integrations/integrations.broadcasters/service-users.js
@@ -28,7 +28,17 @@ export const broadcastServiceUser = async (userId, updateType, roleName, caseRef
 		return false;
 	}
 
-	const msg = mapServiceUserEntity(user, roleName, caseReference);
+	let userAddress;
+	if (user.addressId) {
+		userAddress = await databaseConnector.address.findUnique({
+			where: { id: user.addressId }
+		});
+		if (!userAddress) {
+			userAddress = undefined;
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


appeals/api/src/server/mappers/integration/commands/representation.mapper.js
```diff
@@ -26,7 +26,7 @@ export const mapRepresentationIn = (submission) => {
```
edpackard-work, line 26: not sure if I should map address here instead of updating mapServiceUserIn function, i.e.

```
if (newUser !== null) {
  serviceUser = {
     ...mapServiceUserIn(newUser),
     address: { create: mapAddressIn(newUser) }
  }
}
```

appeals/api/src/server/mappers/integration/commands/service-user.mapper.js
```diff
@@ -14,7 +16,8 @@ export const mapServiceUserIn = (data) => {
 			webAddress: data.website,
 			phoneNumber: data.telephoneNumber,
 			otherPhoneNumber: data.otherPhoneNumber,
-			faxNumber: data.faxNumber
+			faxNumber: data.faxNumber,
+			address: { create: mapAddressIn(data) }
```
edpackard-work, line 20: see comment above - will this break other uses of service user mapper outside representations?

appeals/api/src/server/mappers/integration/commands/representation.mapper.js
fabfab-kc: your call, either is fine from my POV

appeals/api/src/server/mappers/integration/commands/representation.mapper.js
```diff
@@ -26,7 +26,7 @@ export const mapRepresentationIn = (submission) => {
 
 	const attachments = documents.map((doc) => mapDocumentIn(doc, 'representation'));
 	const serviceUserId = data.serviceUserId
-		? Number(data.serviceUserId) - serviceUserIdStartRange
+		? serviceUserIdStartRange - Number(data.serviceUserId)
```
fabfab-kc, line 29: Apologies, this was right the first time... In this case we would get the id as broadcasted, so we need to remove it from the start range...

appeals/api/src/server/mappers/integration/commands/service-user.mapper.js
edpackard-work: it was breaking some other tests where it was being used so have split it into own function

appeals/api/src/server/endpoints/integrations/integrations.broadcasters/service-users.js
```diff
@@ -28,7 +28,17 @@ export const broadcastServiceUser = async (userId, updateType, roleName, caseRef
 		return false;
 	}
 
-	const msg = mapServiceUserEntity(user, roleName, caseReference);
+	let userAddress;
+	if (user.addressId) {
```
fabfab-kc, line 32: probably you can skip this code if you include the address in the DB query on line 22... e.g.:

const user = await databaseConnector.serviceUser.findUnique({
		where: { id: userId },
		include: { address: true }
	});

appeals/api/src/server/mappers/integration/map-service-user-entity.js
```diff
@@ -10,11 +11,12 @@ export const serviceUserIdStartRange = 200000000;
 /**
  *
  * @param {ServiceUser} data
+ * @param {Address | undefined} address
  * @param {string} serviceUserType
  * @param {string} caseReference
  * @returns {AppealServiceUser | undefined}
  */
-export const mapServiceUserEntity = (data, serviceUserType, caseReference) => {
+export const mapServiceUserEntity = (data, address, serviceUserType, caseReference) => {
```
fabfab-kc, line 19: again this is probably redundant if the address is returned with `data`

appeals/api/src/server/mappers/integration/commands/service-user.mapper.js
```diff
@@ -20,3 +22,18 @@ export const mapServiceUserIn = (data) => {
...
+ *
+ * @param {*} data
+ * @returns {import('#db-client').Prisma.ServiceUserCreateInput|undefined}
+ */
+export const mapServiceUserWithAddressIn = (data) => {
+	if (data) {
+		const serviceUser = mapServiceUserIn(data);
+		return {
+			...serviceUser,
+			address: { create: mapAddressIn(data) }
```
fabfab-kc, line 36: would this fail if there's no address?

fabfab-kc (approved):
LGTM!

fabfab-kc:
Hey Ed, looks good - just a couple of things sorry...
I have another PR using model 1.9.19 (latest). 
Very minor... do we need to pass the boolean to the service user mapper? I would have thought just checking for the data would be enough.

edpackard-work:
> Hey Ed, looks good - just a couple of things sorry... I have another PR using model 1.9.19 (latest). Very minor... do we need to pass the boolean to the service user mapper? I would have thought just checking for the data would be enough.



> Hey Ed, looks good - just a couple of things sorry... I have another PR using model 1.9.19 (latest). Very minor... do we need to pass the boolean to the service user mapper? I would have thought just checking for the data would be enough.

it might be overkill, but the mapper is used elsewhere for appellant and agent where an address is created separately and not as part of service user - this is to avoid potential issues with that, though it might not be an issue in practice

fabfab-kc (approved):
LGTM!

PR #1069, opened 2025-03-12
fix(api): re-enable api typechecking

Si3rr4wow:
## Describe your changes

Fixes the last of the type errors and turns API type checking back on 🥳 

[Ticket](https://pins-ds.atlassian.net/browse/A2-1362)

### DIFF
--- a/appeals/api/src/server/endpoints/documents/documents.validators.js
+++ b/appeals/api/src/server/endpoints/documents/documents.validators.js
@@ -19,7 +19,7 @@ import { APPEAL_VIRUS_CHECK_STATUS } from 'pins-data-model';
 import { validateFileNameParameter } from '#common/validators/filename-parameter.js';
 
 /** @typedef {import('@pins/appeals.api').Appeals.UpdateDocumentsRequest} UpdateDocumentsRequest */
-/** @typedef {import('@pins/appeals.api').Appeals.UpdateDocumentsAvCheckRequest} UpdateDocumentsAvCheckRequest */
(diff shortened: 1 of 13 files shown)


appeals/api/src/server/endpoints/documents/documents.validators.js
```diff
@@ -19,7 +19,7 @@ import { APPEAL_VIRUS_CHECK_STATUS } from 'pins-data-model';
 import { validateFileNameParameter } from '#common/validators/filename-parameter.js';
 
 /** @typedef {import('@pins/appeals.api').Appeals.UpdateDocumentsRequest} UpdateDocumentsRequest */
-/** @typedef {import('@pins/appeals.api').Appeals.UpdateDocumentsAvCheckRequest} UpdateDocumentsAvCheckRequest */
+/** @typedef {import('../../openapi-types').UpdateDocumentsAvCheckRequest} UpdateDocumentsAvCheckRequest */
```
fabfab-kc, line 22: probably we should call these references from `import('@pins/appeals.api').Api...` which should be the same but reusable in web...

fabfab-kc (approved):
LGTM!

appeals/api/src/server/endpoints/documents/documents.validators.js
Si3rr4wow: Ah, good pointer, I should've dug more into that, fixing now!

fabfab-kc (approved):
LGTM!

PR #1115, opened 2025-03-20
fix(api): caseValidationDate set on invalid appeals

fabfab-kc:
When an appeal is INVALID, it returns the correct `caseValidationDate` and applies the correct logic based on the state transition history.

### DIFF
--- a/appeals/api/src/server/mappers/integration/shared/map-case-dates.js
+++ b/appeals/api/src/server/mappers/integration/shared/map-case-dates.js
@@ -1,5 +1,6 @@
 import { APPEAL_CASE_STATUS } from 'pins-data-model';
 import { mapDate, findStatusDate } from '#utils/mapping/map-dates.js';
+import { isCaseInvalid } from '#utils/case-invalid.js';
 
 /** @typedef {import('@pins/appeals.api').Schema.Appeal} Appeal */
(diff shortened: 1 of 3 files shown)


appeals/api/src/server/mappers/integration/shared/map-case-dates.js
```diff
@@ -13,6 +14,11 @@ import { mapDate, findStatusDate } from '#utils/mapping/map-dates.js';
 export const mapCaseDates = (data) => {
 	const { appeal } = data;
 
+	const appellantCaseValidationDate = isCaseInvalid(appeal.appealStatus)
+		? findStatusDate(appeal.appealStatus, APPEAL_CASE_STATUS.READY_TO_START)
```
bryn-pins, line 18: this confuses me but that might just be a lack of understanding of the status history from my end

bryn-pins (approved):

appeals/api/src/server/mappers/integration/shared/map-case-dates.js
fabfab-kc: yeah probably not very well named, but basically checks is INVALID because of a decision, rather than the rejection of the appellant case

PR #1116, opened 2025-03-21
fix(web): set supporting document date to today

ShareMyWebStuff:
## Describe your changes

When adding a supporting document to a statement, ip comment or final comment. The submitted date input boxes are blank. This release sets them to todays date.

[Ticket](https://pins-ds.atlassian.net/browse/A2-2635)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/representations/document-attachments/add-document.mapper.js
+++ b/appeals/web/src/server/appeals/appeal-details/representations/document-attachments/add-document.mapper.js
@@ -1,5 +1,5 @@
 import { appealShortReference } from '#lib/appeals-formatter.js';
-import { dateISOStringToDayMonthYearHourMinute } from '#lib/dates.js';
+import { dateISOStringToDayMonthYearHourMinute, getTodaysISOString } from '#lib/dates.js';
 import { dateInput } from '#lib/mappers/index.js';
 
 /** @typedef {import("../../appeal-details.types.js").WebAppeal} Appeal */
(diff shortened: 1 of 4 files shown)


BenSurgisonPINS (dismissed):
lgtm

appeals/web/src/server/appeals/appeal-details/representations/document-attachments/add-document.mapper.js
```diff
@@ -20,7 +21,10 @@ export const dateSubmitted = (appealDetails, errors, date, backLinkUrl) => ({
 		dateInput({
 			id: 'date',
 			name: 'date',
-			value: date,
+			value:
+				date.day && date.month && date.year
+					? date
+					: dateISOStringToDayMonthYearHourMinute(new Date().toISOString()),
```
kandcdev, line 27: we've had a few issues in the past with custom dates, so as a rule of thumb we should not use `new Date()` but rather use lib/dates methods as much as possible

There is a method in `#lib/dates.js` - `getTodaysISOString()`, so this would become `dateISOStringToDisplayDate(getTodaysISOString())`

appeals/web/src/server/appeals/appeal-details/representations/interested-party-comments/add-ip-comment/add-ip-comment.mapper.js
```diff
@@ -230,7 +231,10 @@ export const dateSubmitted = (appealDetails, errors, date, backLinkUrl) => ({
 		dateInput({
 			id: 'date',
 			name: 'date',
-			value: date,
+			value:
+				date.day && date.month && date.year
+					? date
+					: dateISOStringToDayMonthYearHourMinute(new Date().toISOString()),
```
kandcdev, line 237: same as above

kandcdev (approved):

PR #1118, opened 2025-03-21
fix(appeals): dates in lpa and appellant email don't match up (a2-2663)

BenSurgisonPINS:
#### Add missing deadlines to the start case appellant notify (a2-2578)

Please note that any variables added to notify are ignored if they are not used.  It is safe to add the missing variables prior to the template/s being updated  to include them.

Please also note this needs to be deployed to DEV and TEST before the relevant template is updated to include them.  Note also that the duplicate appeal needs to be removed in the template itself.

#### API:
- Add missing lpa_statement_deadline and ip_comments_deadline to the variables for notify

#### TEST:
- All unit tests pass
- Tested within browser

[Ticket: API - Dates in LPA AND Appellant email don't match up to BO (A2-2663)](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2663)


james-hill-kc (approved):
LGTM!

PR #1131, opened 2025-03-25
fix(web): bugfix/A2-2712-fix-LPA-statement-broken-CYA-links

kandcdev:
https://pins-ds.atlassian.net/browse/A2-2712

Fix LPA statements CYA broken links


toyin-pins (approved):
looks good to me!

james-hill-kc (approved):
Approved with a couple of comments, very minor so it's up to you whether you think they are worth addressing

PR #1133, opened 2025-03-25
fix(web): fix ip comment manage folder page back link url

edpackard-work:
* add `/view` to backlink url when user is on interested-party-comments manage documents page

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2691)


glpins (approved):

james-hill-kc (approved):
I'd normally be against adding special-case code to shared functions, but in this case we're looking at refactoring this anyway to resolve this issue in a more robust way, so happy to approve this as an interim fix (this is the tech debt ticket for the proposed refactor https://pins-ds.atlassian.net/browse/A2-2742)

PR #1134, opened 2025-03-25
feat(appeals): case timetable showing rows and action links pre case start (a2-2673)

BenSurgisonPINS:
#### WEB:
- Timetable rows are filtered out via the sub-mappers returning the display parameter as an empty object when they should not be shown.  Also satisfies typescript.
- Added a class to the timetable summary list to allow for snapshots of the whole timetable in unit tests
- Unit tests to test for each AC
- Unit tests and snapshots updated to match

#### TEST:
- All unit tests pass
- Tested within browser

[Ticket:  Case timetable showing rows and action links pre case start (A2-2673)](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2673)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/__tests__/appeal-details.test.js
+++ b/appeals/web/src/server/appeals/appeal-details/__tests__/appeal-details.test.js
@@ -1723,7 +1723,7 @@ describe('appeal-details', () => {
 
 			const unprettifiedElement = parseHtml(response.text, { skipPrettyPrint: true });
 			expect(unprettifiedElement.innerHTML).toContain(
-				'Start date</dt><dd class="govuk-summary-list__value"> Not added</dd>'
+				'Valid date</dt><dd class="govuk-summary-list__value"> 23 May 2023</dd>'
 			);
 		});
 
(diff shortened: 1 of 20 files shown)


BenSurgisonPINS:
> Looks good to me, we should add unit tests to cover the updated mappers if possible (i.e. case shows X when not started and Y when started)

Unit tests added as suggested

appeals/web/src/server/appeals/appeal-details/__tests__/appeal-details.test.js
```diff
@@ -2490,11 +2490,221 @@ describe('appeal-details', () => {
...
+							...appealDataFullPlanning,
+							validAt: null,
+							startedAt: null,
+							appealId
+						});
+					const response = await request.get(`${baseUrl}/${appealId}`);
+					expect(response.statusCode).toBe(200);
+
+					const validDateRow = parseHtml(response.text, {
+						rootElement: `.appeal-valid-date`,
```
james-hill-kc, line 2530: I think this would need to be a selector for the whole timetable (not just the valid date row), otherwise the `.not.toContain` checks further down will always pass regardless of what other items are in the timetable. I'm not sure such a selector exists - if it's not straightforward to add one, you could use the `appeal-valid-date` selector to parse the HTML for the snapshot, and add a separate call to `parseHtml` without it, to be used in the `.not.toContain` checks.

BenSurgisonPINS: I have added a class "appeal-case-timetable" to the whole Timetable summary list.  I have changed the snapshots to use this so the .not.toContain tests are no longer necessary.  For additional belts and braces, I have added separate unit tests for each submapper row to test each AC.  Hopefully I've covered everything.

fabfab-kc (approved):
LGTM!

james-hill-kc (approved):
LGTM!

PR #1135, opened 2025-03-26
fix(api): A2-2615 Wrong Email sent when ip comment is rejected#1097

glpins:
## Describe your changes

Minor tweak to send reject email to appellant email address .

[A2-2615](https://pins-ds.atlassian.net/browse/A2-2615)

[A2-2615]: https://pins-ds.atlassian.net/browse/A2-2615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/endpoints/representations/notify/services/index.js
+++ b/appeals/api/src/server/endpoints/representations/notify/services/index.js
@@ -35,26 +35,24 @@ export const ipCommentRejection = async ({
 	const siteAddress = formatSiteAddress(appeal);
 	const reasons = formatReasons(representation);
 	const extendedDeadline = await formatExtendedDeadline(allowResubmit);
-	const recipientEmail = appeal.agent?.email || appeal.appellant?.email;
-	if (!recipientEmail) {
-		throw new Error(`no recipient email address found for Appeal: ${appeal.reference}`);
-	}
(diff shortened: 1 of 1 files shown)


appeals/api/src/server/endpoints/representations/notify/services/index.js
```diff
@@ -35,7 +35,7 @@ export const ipCommentRejection = async ({
 	const siteAddress = formatSiteAddress(appeal);
 	const reasons = formatReasons(representation);
 	const extendedDeadline = await formatExtendedDeadline(allowResubmit);
-	const recipientEmail = appeal.agent?.email || appeal.appellant?.email;
+	const recipientEmail = appeal.appellant?.email;
```
fabfab-kc, line 38: This should be the `representation.represented?.email`? It sounds like the email is for the user that has submitted the IP comments, rather than the appellant?

fabfab-kc (changes requested):
Unless I misunderstood, I think the email recipient should be the user that has created the ip comment representation, so `representation.represented`...

fabfab-kc (approved):
LGTM!

PR #1140, opened 2025-03-26
fix(web): bugfix/A2-2713-lpa-stmt-cya-allocation-specialisms

kandcdev:
This PR fixes logic for forced allocation answers on CYA page of lpa statement review journey

https://pins-ds.atlassian.net/browse/A2-2712
https://pins-ds.atlassian.net/browse/A2-2713

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/allocation/allocation.mapper.js
+++ b/appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/allocation/allocation.mapper.js
@@ -80,7 +80,7 @@ export function allocationLevelPage(
 	return {
 		title: 'Allocation level',
 		backLinkUrl: `/appeals-service/appeal-details/${appealDetails.appealId}/lpa-statement/${
-			sessionData.forcedAllocation ? '' : `${flowRoute}/allocation-check`
+			sessionData?.forcedAllocation ? '' : `${flowRoute}/allocation-check`
 		}`,
 		preHeading: `Appeal ${shortReference}`,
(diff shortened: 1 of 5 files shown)


appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/allocation/allocation.mapper.js
```diff
@@ -80,7 +80,7 @@ export function allocationLevelPage(
 	return {
 		title: 'Allocation level',
 		backLinkUrl: `/appeals-service/appeal-details/${appealDetails.appealId}/lpa-statement/${
-			sessionData.forcedAllocation ? '' : `${flowRoute}/allocation-check`
+			sessionData?.forcedAllocation ? '' : `${flowRoute}/allocation-check`
```
fabfab-kc, line 83: I wonder why we don't store this by default as `sessionData.${appealDetails.appealId}.xxx` by default, to avoid cross-case leaks? But it's more a general question :-)

glpins (approved):

fabfab-kc (approved):
A small generic comment, probably to discuss with @james-hill-kc 
LGTM!

appeals/web/src/server/appeals/appeal-details/representations/lpa-statement/allocation/allocation.mapper.js
james-hill-kc: It's a good point, ideally we should always be scoping this type of session data by appeal id (plus any other relevant ids eg. representation id)

PR #1144, opened 2025-03-27
fix(web): fix logic to determine if IP comments

edpackard-work:
[Ticket]()


glpins (approved):
Does this not break any snapshots or tests?

PR #1146, opened 2025-03-27
chore(appeals): spike to improve gov notify template management (a2-2762)

BenSurgisonPINS:
#### Notify:
- Added new [generic template](https://www.notifications.service.gov.uk/services/c46d894e-d10e-4c46-a467-019576cd906a/templates/3fd5fc42-d77f-42c2-984a-cf9b89e4c415) in notify test

#### API:
- Updated `.env.example` with new `GOV_NOTIFY_APPEAL_GENERIC_ID` env variable
- Create new notify folder structure including two notify mark down templates for  IP comment rejection emails
- Add notify import path to both `package.json` and `jsconfig.js`
- Created a generic `notifySend` function to populate a supplied template and call `notifyClient.sendEmail`
- Implemented usage of above when sending IP comment rejection emails

#### TEST:
- All unit tests pass
- Tested within browser
- Checked that notify generated the email

[Ticket:  Spike: improve gov notify template management (A2-2762)](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2762)


appeals/api/src/server/notify/notify-send.js
```diff
@@ -0,0 +1,70 @@
...
+ * @property {string} subjectTemplate
+ * @property {import('#endpoints/appeals.js').NotifyClient} notifyClient
+ * @property {string} recipientEmail
+ * @property {Record<string, string | string[]>} personalisation
+ */
+
+/**
+ * @type {Record<string, string>}
+ */
+const templateCache = {};
```
fabfab-kc, line 22: what's the lifetime of this one? should it plug to redis?

BenSurgisonPINS: I'm pretty sure it exists while the instance is running.  When it is restarted they are cached after the first time they are used.  If there are multiple instances of the application in a cluster for example, each one would have it's own cache.  It just saves reading from the file every time a template is required.

BenSurgisonPINS: We could use redis as it may save a little bit of memory and prevent some of the reads from the files, but I'm not sure we would save much.

BenSurgisonPINS:
Moved spike code to actual ticket https://pins-ds.atlassian.net.mcas.ms/browse/A2-2732

PR #1147, opened 2025-03-27
fix(appeals): correcting final comment text displayed in case history (a2-2687)

pinsMT:
- Changed displayed text to "appellanbt/lpa final comment accpeted"
- Changed other final comments text to uniform sentence  
- Added IF condition for accepted comment status to determine correct text for case history update

https://pins-ds.atlassian.net/browse/A2-2687


fabfab-kc (approved):
LGTM!

PR #1153, opened 2025-03-28
fix/a2-2411-correcting-documentation-final-comments-fields

pinsMT:
- Updated null text to "No final comments"
- Set null date text to "Not applicable"

https://pins-ds.atlassian.net/browse/A2-2411


BenSurgisonPINS (approved):
lgtm

PR #1161, opened 2025-03-31
feat(web): add s78 start case select procedure type page and entry points

james-hill-kc:
## Describe your changes

Adds a new page to start case flow (s78 select procedure type) and entry points, including new functionality to pass the back link URL using a query string (based on existing approach used elsewhere in the codebase), plus some unit tests.

https://pins-ds.atlassian.net/browse/A2-2636


fabfab-kc (approved):
LGTM!

PR #1170, opened 2025-04-02
feat(appeals): emulate notify in local dev (a2-2732)

BenSurgisonPINS:
[Ticket]()


appeals/api/src/server/notify/emulate-notify.js
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	const emailHtml = [
+		`<b>Recipient email:</b> <span>${recipientEmail}</span>`,
+		`<b>Subject:</b> ${subject}`,
+		'<hr>',
+		contentLines.join('\n')
+	].join('<br>\n');
+
+	const outputDir = path.join(process.cwd(), 'temp');
```
fabfab-kc, line 61: Wondering if this is created automatically and if we should add its contents to .gitignore

BenSurgisonPINS: Don't worry, all temp folders and their contents are ignored as it's already in .gitignore

PR #1179, opened 2025-04-03
fix(web): quick fix for inability to start s78 cases due to select procedure page

james-hill-kc:
## Describe your changes

Quick fix to allow s78 cases to be started despite POST functionality for the select procedure type not being in place yet

https://pins-ds.atlassian.net/browse/A2-2954


fabfab-kc (approved):
LGTM!

PR #1180, opened 2025-04-03
fix: use worker limits and log heap

bryn-pins:
## Describe your changes

Uses worker limits to reduce memory usage

https://jestjs.io/docs/configuration#workeridlememorylimit-numberstring


fabfab-kc (approved):
Keen to try... my review may be pointless if no permissions!

PR #1181, opened 2025-04-03
feat(appeals): hearing check details and start case page

james-hill-kc:
## Describe your changes

Web and API code to implement check details and start case page and related actions, when user selects "hearing" as the procedure type when starting an appeal.

https://pins-ds.atlassian.net/browse/A2-2667


appeals/api/src/server/endpoints/appeal-timetables/appeal-timetables.service.js
```diff
@@ -55,9 +55,17 @@ const checkAppealTimetableExists = async (req, res, next) => {
...
  * @returns
  */
-const startCase = async (appeal, startDate, notifyClient, siteAddress, azureAdUserId) => {
+const startCase = async (
+	appeal,
+	startDate,
+	notifyClient,
+	siteAddress,
+	azureAdUserId,
+	procedureType
```
fabfab-kc, line 67: I would probably make this nullable (defaults to 'written')... It depends if we are starting HAS cases the same way...

fabfab-kc (dismissed):
LGTM! I would only pay attention that the HAS cases start correctly (if the procedure is not passed from the FE)...

appeals/api/src/server/endpoints/appeal-timetables/appeal-timetables.service.js
james-hill-kc: The HAS cases do use the same code, but `procedureType` is an optional parameter and will only be passed to `appealRepository.updateAppealById` if it's provided. I've tested starting a HAS case (which doesn't pass the procedureType from the web app) and it seems to work fine. Is there any other reason this would need to be nullable?

fabfab-kc (approved):

appeals/web/package.json
```diff
@@ -19,7 +19,7 @@
     "rollup": "node ./scripts/rollup/bundle-js.js",
     "sass": "node ./scripts/rollup/compile-css.js",
     "start": "node ./src/server/server.js",
-    "test": "cross-env NODE_ENV=test NODE_OPTIONS=\"--experimental-vm-modules --expose-gc\" npx jest --logHeapUsage --ci",
+    "test": "cross-env NODE_ENV=test NODE_OPTIONS=\"--experimental-vm-modules --expose-gc\" npx jest -u --logHeapUsage --ci",
```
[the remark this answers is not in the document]
james-hill-kc: removed, thanks!

PR #1186, opened 2025-04-07
fix(web): changing green button to grey, adding a div to make text inline with span

Callum-T-Dev:
Updated snapshots for amended button

[2787](https://pins-ds.atlassian.net/jira/software/c/projects/A2/boards/246/backlog?selectedIssue=A2-2787)


james-hill-kc (approved):
Looks good to me!

PR #1193, opened 2025-04-08
chore(api): improves SQL batch performance, upgrades to prisma 6.5.0

fabfab-kc:
- Upgrades `prisma` and `prisma-client` to v6.5.0
- Adds a `QUERY_BATCH_SIZE = 2000` API env entry, to handle `selectMany` with lots of data parameters
- Adds a DB index on all relationships that are queried with filters


kandcdev (approved):

BenSurgisonPINS (approved):
LGTM

PR #1194, opened 2025-04-08
A2-2979(fix): updating show more logic to highlight text

toyin-pins:
Fixing issue where show-more component was not initialised when it was needed. 
- Updated logic so that if rowSelector is null (i.e. contentRowSelector has not been specified) which means we have not said what should be used to define individual rows, we now check the number of characters as we do for text mode
- Updated the css for the show-more component so that highlight applies even when the component is not fully initialised
[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2979)

### DIFF
--- a/appeals/web/src/styles/7-components/show-more.scss
+++ b/appeals/web/src/styles/7-components/show-more.scss
@@ -13,10 +13,8 @@
 		text-decoration: underline;
 	}
 
-	&__content {
-		mark {
-			background-color: #f4cdc6
-		}
+	mark {
+		background-color: #f4cdc6
 	}
 }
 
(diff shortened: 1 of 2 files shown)


appeals/web/src/styles/7-components/show-more.scss
```diff
@@ -13,10 +13,8 @@
 		text-decoration: underline;
 	}
 
-	&__content {
-		mark {
-			background-color: #f4cdc6
-		}
+	mark {
```
james-hill-kc, line 16: Is this overriding a default GDS style? Just wondering if removing this entirely would also fix the issue in a more global way

james-hill-kc (approved):
LGTM! I left one comment but it's just a question

PR #1195, opened 2025-04-08
fix/a2-2392-lpa-reivew-statement-text-width-correction

pinsMT:
-Removed HTML wrappers
-Set display classes to "no actions" to allow width expansion
-Added new njk files
-imported new njk files into pattern display html

https://pins-ds.atlassian.net/browse/A2-2392


james-hill-kc (approved):
LGTM!

PR #1198, opened 2025-04-09
fix(web): A2-2241 s78 - development types row missing or not displayed correctly

glpins:
## Describe your changes

Added Development Type change flow to Appellant Case page; fixed API bug preventing PATCH responses from returning updated developmentType

[A2-2241](https://pins-ds.atlassian.net/browse/A2-2241)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/appellant-case/application-development-type/application-development-type.service.js
+++ b/appeals/web/src/server/appeals/appeal-details/appellant-case/application-development-type/application-development-type.service.js
@@ -1,0 +1,14 @@
+/**
+ * @param {import('got').Got} apiClient
+ * @param {string} appealId
+ * @param {string} appellantCaseId
+ * @param {string | null} developmentType
+ * @returns {Promise<{}>}
+ */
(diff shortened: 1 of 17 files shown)


fabfab-kc (approved):
LGTM!

appeals/web/src/server/appeals/appeal-details/appellant-case/application-development-type/application-development-type.service.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @param {import('got').Got} apiClient
+ * @param {string} appealId
+ * @param {string} appellantCaseId
+ * @param {string | null} developmentType
+ * @returns {Promise<{}>}
+ */
+export function changeDevelopmentType(apiClient, appealId, appellantCaseId, developmentType) {
```
toyin-pins, line 8: very minor but you've tended toward arrow notation when defining functions everywhere else, might be worth being consistent

glpins: Plan is to leave for now to keep consistency with the other component services

PR #1208, opened 2025-04-10
feat(web): global back link fixes - flow entry points

james-hill-kc:
## Describe your changes

- Adds helper functions `addBackLinkQueryToUrl` and `getBackLinkUrlFromQuery` to support easily adding and retrieving a uniform `backUrl` query to any href, to allow for implementing arbitrary back links where required
- Applies proper back link behaviour (using the helpers mentioned above) to all flow entry points (from personal list, case details important banners, and case details row action links) 

https://pins-ds.atlassian.net/browse/A2-2945

### DIFF
--- a/appeals/api/src/database/seed/data-test.js
+++ b/appeals/api/src/database/seed/data-test.js
@@ -399,6 +399,51 @@ const newS78Appeals = [
 	}),
 	appealFactory({
 		typeShorthand: APPEAL_TYPE_SHORTHAND_FPA,
+		status: { status: APPEAL_CASE_STATUS.STATEMENTS, createdAt: getPastDate({ months: 6 }) },
+		lpaQuestionnaire: true,
+		startedAt: getPastDate({ months: 2 }),
+		validAt: getPastDate({ months: 6 }),
+		assignCaseOfficer: true,
+		agent: false
+	}),
+	appealFactory({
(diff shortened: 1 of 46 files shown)


appeals/web/src/server/lib/back-link-url.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @param {import('@pins/express/types/express.js').Request} request
+ * @param {string} url
+ * @returns {string}
+ */
+export function addBackLinkQueryToUrl(request, url) {
+	return `${url}?backUrl=${request.originalUrl}`;
```
kandcdev, line 7: one thing to note, currently `backUrl` are using `constructUrl` from `appeals/web/src/server/lib/mappers/utils/url.mapper.js`.
Most of the appeal routes are prefixed with `/appeals-service/appeal-details/${appealId}`. We don't have to pass this part of the string - it'll be prefixed in there. 

Example: `?backUrl=/share` would direct a user to `/appeals-service/appeal-details/${appealId}/share`.

Hope it makes sense

james-hill-kc: Thanks Yan, I'd expect this work will eventually replace `constructUrl` entirely, but it's worth keeping in mind!

kandcdev (approved):

PR #1212, opened 2025-04-11
A2-2078(fix): Uncommitted files persisting when moving out of flow

toyin-pins:
## Describe your changes
- Added common middleware to remove uncommitted files from session
- Applying middleware function to top-level pages GETs to ensure session is clear of uncommitted files

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2078)


BenSurgisonPINS (approved):
LGTM

PR #1217, opened 2025-04-14
fix(web): a2-3031 change case for current version tag

edpackard-work:
* replaces CURRENT VERSION with Current version on the tag the document version history - snapshots updated accordingly
 

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3031)


fabfab-kc (approved):
LGTM!

PR #1226, opened 2025-04-16
feat(api): notifications logger

fabfab-kc:
Notification logger and api endpoint to retrieve emails sent from the system.
It will start collecting data on the new notification internal templates.

[Ticket](https://pins-ds.atlassian.net/browse/A2-2741)

### DIFF
--- a/appeals/api/src/server/notify/notify-send.js
+++ b/appeals/api/src/server/notify/notify-send.js
@@ -5,7 +5,8 @@ import config from '#config/config.js';
 import {
 	ERROR_FAILED_TO_POPULATE_NOTIFICATION_EMAIL,
 	ERROR_FAILED_TO_SEND_NOTIFICATION_EMAIL,
-	ERROR_NO_RECIPIENT_EMAIL
+	ERROR_NO_RECIPIENT_EMAIL,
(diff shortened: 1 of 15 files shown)


appeals/api/src/server/notify/notify-send.js
```diff
@@ -133,6 +135,9 @@ export const notifySend = async (options) => {
 	if (!recipientEmail) {
 		throw new Error(ERROR_NO_RECIPIENT_EMAIL);
 	}
+	if (!personalisation || personalisation.appeal_reference_number === null) {
```
BenSurgisonPINS, line 138: Should we be checking for anything falsey such as an empty string or undefined as well as null?

BenSurgisonPINS (commented):
Just one comment

appeals/api/src/server/notify/notify-send.js
fabfab-kc: sure... need to fix the tests anyways...

glpins (dismissed):
LGTM

glpins (approved):

ShareMyWebStuff (approved):

PR #1228, opened 2025-04-16
feat(web): hide extension when changing filenames

jimbali:
## Describe your changes

- The file extension is now hidden when changing the filename of a document
- New versions of a document must have the same file extension as the original document

[A2-2926](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2926?McasTsid=20596&McasCtx=4)

[A2-2926]: https://pins-ds.atlassian.net/browse/A2-2926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/web/src/server/appeals/appeal-documents/__tests__/appeal-documents.test.js
+++ b/appeals/web/src/server/appeals/appeal-documents/__tests__/appeal-documents.test.js
@@ -246,13 +246,13 @@ describe('appeal-documents', () => {
 
 			expect(element).toContain('Change document details');
 			expect(element).toContain(fileInfo.latestDocumentVersion.originalFilename);
-			expect(element).toContain(fileInfo.name);
+			expect(element).toContain(fileInfo.name.substring(0, fileInfo.name.lastIndexOf('.')));
 		});
 
 		it('should redirect to manage documents page after change document name success', async () => {
 			const response = await request
 				.post(fullUrl)
-				.send({ fileName: 'valid-fileName_123.jpg', documentId });
+				.send({ fileName: 'valid-fileName_123', documentId });
 			expect(response.statusCode).toBe(302);
 			expect(response.text).toContain(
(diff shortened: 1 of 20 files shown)


appeals/web/src/server/appeals/appeal-documents/__tests__/appeal-documents.test.js
```diff
@@ -293,13 +293,13 @@ describe('appeal-documents', () => {
 			}).innerHTML;
 
 			expect(element).toContain(
-				'>File name must only include letters a to z, numbers 0 to 9 and special characters such as hyphens, underscores and one full stop'
+				'>File name must only include letters a to z, numbers 0 to 9 and special characters such as hyphens and underscores'
 			);
 		});
 
 		it('should render change filename page with duplicate filename error', async () => {
```
toyin-pins, line 300: I feel like there's a potential test case missing here where the file extension don't match?

toyin-pins (changes requested):

BenSurgisonPINS (approved):
LGTM (just one observation)

appeals/web/src/server/appeals/appeal-documents/__tests__/appeal-documents.test.js
jimbali: The validation for making sure the extensions match is for uploading a new version of a file, and the validation for that happens on the client side. I could add some additional validation as a middleware in the web app or in the API, but the other validations for that page just seem to happen on the client side.

toyin-pins: ahh yes you're absolutely right. and since it's client side we don't really have tests for this. happy enough

toyin-pins (approved):

PR #1229, opened 2025-04-16
fix(web): remove check for if startedAt is in the past

HadrianHughesKC:
## Describe your changes

* fix(web): remove check for if startedAt is in the past


BenSurgisonPINS (approved):
LGTM

PR #1231, opened 2025-04-17
fix(web): bugfix/A2-2499-lpaq-timetable-change-link

kandcdev:
https://pins-ds.atlassian.net/browse/A2-2499


BenSurgisonPINS (approved):
LGTM

PR #1264, opened 2025-04-25
feat(web): add hearing estimates page

jimbali:
## Describe your changes

First page of add hearing estimates user journey

[A2-2749](https://pins-ds.atlassian.net.mcas.ms/browse/A2-2749?McasTsid=20596&McasCtx=4)

[A2-2749]: https://pins-ds.atlassian.net/browse/A2-2749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/hearing/add-estimates/__tests__/add-estimates.test.js
+++ b/appeals/web/src/server/appeals/appeal-details/hearing/add-estimates/__tests__/add-estimates.test.js
@@ -1,0 +1,197 @@
+// @ts-nocheck
+import { appealData } from '#testing/app/fixtures/referencedata.js';
+import { createTestEnvironment } from '#testing/index.js';
+import { parseHtml } from '@pins/platform';
+import nock from 'nock';
+import supertest from 'supertest';
+
+const { app, installMockApi, teardown } = createTestEnvironment();
+const request = supertest(app);
+const baseUrl = '/appeals-service/appeal-details';
+
+/**
+ * Creates validation test cases for a field that allows days or half days
+ * @param {string} fieldName - The name of the field being tested
+ * @param {string} label - The label used in error messages
+ * @param {Object} validValues - The valid values for other fields
+ * @returns {() => void} A function that runs the validation test cases
+ */
+const behavesLikeADayOrHalfDayField = (fieldName, label, validValues) => {
+	return () => {
+		it('should return 400 when field is empty', async () => {
+			const response = await request
+				.post(`${baseUrl}/2/hearing/add-estimates/estimates`)
+				.send({ ...validValues, [fieldName]: '' });
+
+			expect(response.statusCode).toBe(400);
+			expect(response.text).toContain(`Enter estimated ${label}`);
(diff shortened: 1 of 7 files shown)


appeals/web/src/server/appeals/appeal-details/hearing/add-estimates/__tests__/add-estimates.test.js
```diff
@@ -0,0 +1,197 @@
...
+const baseUrl = '/appeals-service/appeal-details';
+
+/**
+ * Creates validation test cases for a field that allows days or half days
+ * @param {string} fieldName - The name of the field being tested
+ * @param {string} label - The label used in error messages
+ * @param {Object} validValues - The valid values for other fields
+ * @returns {() => void} A function that runs the validation test cases
+ */
+const behavesLikeADayOrHalfDayField = (fieldName, label, validValues) => {
```
james-hill-kc, line 18: Just a suggestion, this approach is fine, but it does deviate slightly from the way we're doing this in other tests. For consistency, it might be better to move this into a loop which iterates an array of test cases containing the data that's currently being passed as parameters: https://github.com/Planning-Inspectorate/appeals-back-office/blob/main/appeals/web/src/server/appeals/appeal-details/costs/__tests__/costs.test.js#L402

appeals/web/src/server/appeals/appeal-details/hearing/add-estimates/__tests__/add-estimates.test.js
```diff
@@ -0,0 +1,197 @@
...
+					.expect('Location', `${baseUrl}/2/hearing/add-estimates/estimates`)
+					.end(resolve);
+			});
+		});
+	});
+
+	describe('GET /hearing/add-estimates/estimates', () => {
+		const appealId = 2;
+
+		/** @type {*} */
```
james-hill-kc, line 95: We're ignoring TS entirely for most of the test files, you may want to do that here too, to avoid needing these annotations (just a thought)

james-hill-kc (approved):
LGTM! Approved with a couple of suggestions

appeals/web/src/server/appeals/appeal-details/hearing/add-estimates/__tests__/add-estimates.test.js
jimbali: I get the consistency argument, but tbh I think I prefer this style because it seems more like a function that would be reusable in other test files. For example if there was another page that had day or half day fields then it might use the same validator, and it could also use this same shared examples function as it is complimentary to the validator.

james-hill-kc: That's a fair point! I'd suggest moving it into its own file, but that can wait until there's a second use-case for it

james-hill-kc (approved):

PR #1275, opened 2025-04-28
fix(api): updating api validator to accept max 8000 chars for extra conditions details

toyin-pins:
## Describe your changes
- Small fix to increase the char limit on the API side for extra condition details 

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-1909)


james-hill-kc (approved):
LGTM! Not sure if there is any precedent for unit testing this in the API code, but if so, it would be good to add test coverage

james-hill-kc (approved):
LGTM!

PR #1300, opened 2025-05-06
feat(appeals): notify read templates asynchronously

BenSurgisonPINS:
## Describe your changes

Notify Templates are now loaded asynchronously

All unit tests still pass.
Sending notify email still works.

### DIFF
--- a/appeals/api/src/server/notify/notify-send.js
+++ b/appeals/api/src/server/notify/notify-send.js
@@ -28,6 +28,11 @@ const __dirname = path.dirname(__filename); // get the name of the directory
  */
 
 /**
+ * @type {Record<string, string>}
+ */
+const templateCache = {};
+
+/**
  * Template variable names must:
  * - Be within double brackets (( ... ))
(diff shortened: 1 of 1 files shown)


appeals/api/src/server/notify/notify-send.js
```diff
@@ -66,25 +61,28 @@ function validateTemplate(template) {
  * Retrieves template held in cache or from the file if not retrieved before
  *
  * @param {string} templateName
- * @returns {string}
+ * @returns {Promise<string>}
  */
-function getTemplate(templateName) {
-	if (!templateCache[templateName]) {
```
fabfab-kc, line 72: Hey @BenSurgisonPINS without the cache, should we use redis now maybe? To avoid too many `fs` calls?

BenSurgisonPINS: I've restored the cache

fabfab-kc (approved):
LGTM!

PR #1312, opened 2025-05-07
fix(api): typo in notify decision

fabfab-kc:
Fixes a typo in the appeal decision notification template

image


glpins (approved):

appeals/api/src/server/notify/templates/decision-is-allowed-split-dismissed-appellant.content.md
```diff
@@ -8,13 +8,13 @@ Planning application reference: ((lpa_reference))
 
 We have made a decision on your appeal.
 
-[Sign in to our service](https://appeals-service.planninginspectorate.gov.uk/appeals/((appeal_reference_number))) to view the decision letter dated ((decision_date)).
+[Sign in to our service](<https://appeals-service.planninginspectorate.gov.uk/appeals/((appeal_reference_number))>) to view the decision letter dated ((decision_date)).
```
BenSurgisonPINS, line 11: I don't think we should have these <> brackets

fabfab-kc: yeah struggling to get rid of them as git doesn't detect the change

PR #1317, opened 2025-05-07
Fix/a2 2610 correcting case history representation display name

pinsMT:
- Added value for audit display name
- Added test

Ticket:
https://pins-ds.atlassian.net/browse/A2-2610


appeals/web/src/server/appeals/appeal-details/audit/audit.service.js
```diff
@@ -136,10 +144,16 @@ const tryMapDocument = async (appealId, log, docInfo, lpaqId) => {
...
 
 				const url = `/appeals-service/appeal-details/${appealId}/costs/${path}/manage-documents/${folderId}/${documentGuid}`;
-				return log.replace(name, `<a class="govuk-link" href="${url}">${name}</a>`);
+				return log.replace(
+					repAuditDisplayName,
+					`<a class="govuk-link" href="${url}">${repAuditDisplayName}</a>`
+				);
+			}
+			case 'representation': {
+				return log.replace(name, `<a class="govuk-link" href="#">${repAuditDisplayName}</a>`);
```
fabfab-kc, line 153: If you want to target only representations (where by default there will be a `guid` prepending the name), you only need to use it here, but not sure if we want to do it everywhere...

fabfab-kc (dismissed):
LGTM! One comment around business requirements...

appeals/web/src/server/appeals/appeal-details/audit/audit.service.js
pinsMT: I believe this is the only place we need to apply this logic.

PR #1331, opened 2025-05-09
fix(api): hotfix on dashboard url

fabfab-kc:
Fixes a live issue (404) with the link to the LPA dashboard in a notification template. This branch is created out of our deployed production release

### DIFF
--- a/appeals/api/src/server/notify/templates/decision-is-allowed-split-dismissed-lpa.content.md
+++ b/appeals/api/src/server/notify/templates/decision-is-allowed-split-dismissed-lpa.content.md
@@ -8,7 +8,7 @@ Planning application reference: ((lpa_reference))
 
 A decision has been made on this appeal.
 
-[Sign in to our service](https://appeals-service.planninginspectorate.gov.uk/manage-appeals/((appeal_reference_number))) to view the decision letter dated ((decision_date)).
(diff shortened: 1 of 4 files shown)


appeals/api/src/server/notify/templates/decision-is-allowed-split-dismissed-lpa.content.md
```diff
@@ -8,7 +8,7 @@ Planning application reference: ((lpa_reference))
 
 A decision has been made on this appeal.
 
-[Sign in to our service](https://appeals-service.planninginspectorate.gov.uk/manage-appeals/((appeal_reference_number))) to view the decision letter dated ((decision_date)).
+[Sign in to our service](<https://appeal-planning-decision.service.gov.uk/manage-appeals/((appeal_reference_number))>) to view the decision letter dated ((decision_date)).
```
bryn-pins, line 11: this should be correct now
But we need to update decision-is-allowed-split-dismissed-appellant.content.md as well

Both paths are correct, just the domain is wrong, seems good for a hotfix but would suggest getting these into env vars wherever used

fabfab-kc: yeah we have a ticket to add the env vars... I'll look at the other template and the annoying issue with the < which are failing the tests

bryn-pins (approved):

PR #1345, opened 2025-05-13
fix/a2-2610-adding-redaction-status-to-rep-audit-display-name

pinsMT:
- Amended logic within add documents function to add reaction status into details text for rep audit logs

https://pins-ds.atlassian.net/browse/A2-2610


fabfab-kc (approved):
LGTM!

PR #1349, opened 2025-05-13
fix(api): clear test data not deleting notifications

fabfab-kc:
Fixes the `clear test data` script, which was failing as unable to delete notifications from test cases.

Error details: https://dev.azure.com/planninginspectorate/appeals-back-office/_build/results?buildId=114206&view=logs&j=0a1d8c75-6a3d-5944-5d56-564de451fe0b

### DIFF
--- a/appeals/api/src/database/seed/seed-clear-test-data.js
+++ b/appeals/api/src/database/seed/seed-clear-test-data.js
@@ -8,19 +8,28 @@ import { databaseConnector } from '../../server/utils/database-connector.js';
 async function deleteTestRecords() {
 	const lpaCodes = localPlanningDepartmentList.map((lpa) => lpa.lpaCode);
 
-	const appealIDs = await getAppeals(lpaCodes);
-	if (appealIDs.length === 0) {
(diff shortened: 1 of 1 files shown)


appeals/api/src/database/seed/seed-clear-test-data.js
```diff
@@ -32,13 +41,15 @@ async function deleteTestRecords() {
 /**
  *
  * @param {number[]} appealIDs
+ * @param {string[]} appealRefs
  * @param {number[]} appellantCasesIDs
  * @param {number[]} lpaqIDs
  * @param {string[]} documentIDs
  * @param {number[]} representationIDs
  */
 const deleteAppealData = async (
```
HadrianHughesKC, line 50: Shouldn't be a big deal, but with this interface the caller could pass different lists for appealIDs and appealRefs, which could cause errors. Safer would be to just pass appealIDs and query for the relevant appealRefs inside this function.

HadrianHughesKC (approved):

PR #1354, opened 2025-05-14
fix(api): adjusting FO env URLs for email templates dashboard links

fabfab-kc:
Fixes a conflict on config settings for FO URLs.


james-hill-kc (approved):
LGTM!

PR #1360, opened 2025-05-14
fix(api): notify template for start S78 case LPA - add timetable - A2-3301

fabfab-kc:
Adds the timetable to the email sent to LPA when starting S78 case.
[Ticket](https://pins-ds.atlassian.net/browse/A2-3301)


BenSurgisonPINS (approved):
LGTM

PR #1397, opened 2025-05-20
chore(appeals): issue decision clean-up routes (a2-3162)

BenSurgisonPINS:
### WEB:
- Remove redundant "validateAppeal" middleware as it is already in the parent route.  see appeal-details.router.js
- Move clearing the session cache for the decision types when the appeal-id is different into middleware for every route
- Add some comments describing why the reuse of the appeal-documents upload page functions requires storing and restoring the uploadFileInfo object in the session for each decision type.

### TEST:
- Made sure all unit tests pass

[Issuing decision flow - Part 7 - Check details (A2-3162)](https://pins-ds.atlassian.net/browse/A2-3162)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/issue-decision/issue-decision.router.js
+++ b/appeals/web/src/server/appeals/appeal-details/issue-decision/issue-decision.router.js
@@ -2,11 +2,13 @@ import { Router as createRouter } from 'express';
 import { asyncHandler } from '@pins/express';
 import * as controller from './issue-decision.controller.js';
 import * as validators from './issue-decision.validators.js';
-import { validateAppeal } from '../appeal-details.middleware.js';
 import { assertUserHasPermission } from '#app/auth/auth.guards.js';
 import { permissionNames } from '#environment/permissions.js';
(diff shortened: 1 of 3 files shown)


appeals/web/src/server/appeals/appeal-details/issue-decision/issue-decision.router.js
```diff
@@ -23,12 +25,10 @@ router
 router
 	.route('/decision-letter-upload')
 	.get(
-		validateAppeal,
```
james-hill-kc, line 26: Just checking, do these routes definitely not need this middleware anymore?

james-hill-kc (approved):
LGTM! Just one question but nothing that should block a merge

appeals/web/src/server/appeals/appeal-details/issue-decision/issue-decision.router.js
BenSurgisonPINS: "validateAppeal" middleware is already in the parent route. see appeal-details.router.js

PR #1399, opened 2025-05-20
fix(web): a2-3195 bugfix update change LPA id prefix

edpackard-work:
* update id prefix so error message behaviour is as expected

[Ticket](https://pins-ds.atlassian.net/browse/A2-3195)


BenSurgisonPINS (approved):
LGTM

PR #1422, opened 2025-05-23
feat(appeals): saving lpaq due date to db and updating case history

toyin-pins:
**API**:
- Updated case history timetable update content
-  Creating audit trail for each of the update timetable types separately (only 1 for HAS)

**Web**
- Added a check to see if there has been any change to any of the timetables - if not, skip calling api and just return success banner and redirect user to appeal details page
-  removing unused code
- Copied setTimeInTimeZone method over to web package from api

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3121)


BenSurgisonPINS (approved):
LGTM

PR #1428, opened 2025-05-28
fix(appeals): only bypass duplicate file validation when doc stage is representation (a2-3404)

BenSurgisonPINS:
#### Only by-pass duplicate file validation when the document stage is representation (a2-3404)

Web:
- Moved bypass and duplicate file validation after the other client-side file validation is complete

[Ticket (A2-3404): Can update version to different file type or exceed filesize limits](https://pins-ds.atlassian.net/browse/A2-3404)


james-hill-kc (dismissed):
LGTM!

PR #1460, opened 2025-06-03
fix(appeals): remove unnecessary flakey snapshot due to changing current date

BenSurgisonPINS:
Remove an unnecessary snapshot that keeps changing due to it containing the current date.


jimbali (approved):
Ben to the rescue!

PR #1461, opened 2025-06-03
feat(appeals): updating case history notes formatting for timetable updates

toyin-pins:
## Describe your changes
# API
- Updating the case history auditTrail for timetable updates - every new timetable update gets added to a new line starting with Timetable update

# Web
- Updating the request from the web side so that only the updated timetables get sent to the API side and any that are unchanged get removed from the sessionTimetable
- Small copy update to the success banner message
- Updating some tests 

[Ticket]()


andy-mair121 (approved):
LGTM

PR #1483, opened 2025-06-06
fix(web): removing appeal timetable from session before redirect

toyin-pins:
## Describe your changes
- fixing small bug where session was not cleared when no changes were made to timetables

[Ticket](https://pins-ds.atlassian.net/browse/A2-3097)


BenSurgisonPINS (approved):
lgtm

PR #1496, opened 2025-06-10
Fix/a2 3032 display application type on appellant case

edpackard-work:
* returns the type of application as part of appellant case from API
* displays the type of application on appellant case view, instead of (as previously) appeal type
  

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3032)


fabfab-kc (approved):
LGTM!

PR #1500, opened 2025-06-10
fix/a2-3236-removed-final-comment-rows-from-documenation-table

pinsMT:
- added conditions to mappers which remove appellant or lpa final comment rows for hearing cases
- added tests

https://pins-ds.atlassian.net/browse/A2-3236


jimbali (approved):
lgtm!

PR #1554, opened 2025-06-20
fix(appeals): changed text from no answered to no

peter-uzor-pins:
## Describe your changes

changed text from no answered to no

[A2-3485](https://pins-ds.atlassian.net/browse/A2-3485)

[A2-3485]: https://pins-ds.atlassian.net/browse/A2-3485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/web/src/server/lib/mappers/data/appellant-case/submappers/other-tenants-of-agricultural-holding.js
+++ b/appeals/web/src/server/lib/mappers/data/appellant-case/submappers/other-tenants-of-agricultural-holding.js
@@ -10,7 +10,7 @@ export const mapOtherTenantsOfAgriculturalHolding = ({
 		id: 'other-tenants-of-agricultural-holding',
 		text: 'Are there any other tenants?',
 		value: appellantCaseData.agriculturalHolding?.hasOtherTenants,
-		defaultText: 'No answered',
+		defaultText: 'No',
 		link: `${currentRoute}/agricultural-holding/other-tenants/change`,
 		editable: userHasUpdateCase
 	});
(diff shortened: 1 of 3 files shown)


appeals/web/src/server/lib/mappers/data/appellant-case/submappers/other-tenants-of-agricultural-holding.js
```diff
@@ -10,7 +10,7 @@ export const mapOtherTenantsOfAgriculturalHolding = ({
 		id: 'other-tenants-of-agricultural-holding',
 		text: 'Are there any other tenants?',
 		value: appellantCaseData.agriculturalHolding?.hasOtherTenants,
-		defaultText: 'No answered',
+		defaultText: 'No',
```
jimbali, line 13: Should this definitely by "No" and not "Not answered"? I think this question is only asked if you said yes that it was an agricultural holding.

peter-uzor-pins: According to the ticket - [A2-3485](https://pins-ds.atlassian.net/browse/A2-3485) 

It should be No and not "No answered". I guess they want the default value changed to No instead, hence my change. Yes i think its only asked if its an agricultural holding.

Also i am not sure if the data was previously missing due to the appealcase mapping issue you resolved. But by default according to the ticket, they want it to be Yes or No.

jimbali (approved):

PR #1558, opened 2025-06-20
feat(web): adding appeal form for CAS planning

zoebutterworth:
and also updating the listed building case where relevant now that we have updated the data model

[Ticket](https://pins-ds.atlassian.net/browse/A2-3263)


appeals/web/src/server/appeals/appeal-details/appellant-case/__tests__/appellant-case.test.js
```diff
@@ -219,6 +220,37 @@ describe('appellant-case', () => {
...
+
+			const unprettifiedElement = parseHtml(response.text, { skipPrettyPrint: true });
+
+			expect(unprettifiedElement.innerHTML).toContain('Appellant case</h1>');
+			expect(unprettifiedElement.innerHTML).toContain('1. Appellant details</h2>');
+			expect(unprettifiedElement.innerHTML).toContain('2. Site details</h2>');
+			expect(unprettifiedElement.innerHTML).toContain('3. Application details</h2>');
+			expect(unprettifiedElement.innerHTML).toContain('4. Appeal details</h2>');
+			expect(unprettifiedElement.innerHTML).toContain('5. Upload documents</h2>');
+			expect(unprettifiedElement.innerHTML).not.toContain('Additional documents</h2>');
```
toyin-pins, line 251: Any CAS specific rows for appellant case we can check for here? Happy for you to self-resolve

appeals/web/src/server/lib/mappers/data/appellant-case/cas.js
```diff
@@ -0,0 +1,47 @@
...
+import { mapApplicationType } from './submappers/application-type.js';
+import { mapInspectorAccess } from './submappers/inspector-access.js';
+import { mapHealthAndSafetyIssues } from './submappers/health-and-safety-issues.js';
+import { mapApplicationForm } from './submappers/application-form.js';
+import { mapDecisionLetter } from './submappers/decision-letter.js';
+import { mapAppealStatement } from './submappers/appeal-statement.js';
+import { mapCostsDocument } from './submappers/costs-document.js';
+
+export const submaps = {
+	appellant: mapAppellant,
```
toyin-pins, line 25: I think these are all the same as the HAS submaps. You should replace all of this with just:
export const submaps = {
         ...hasSubmaps,
	additionalDocuments: mapAdditionalDocuments,
};

toyin-pins (changes requested):

appeals/web/src/server/appeals/appeal-details/appellant-case/__tests__/appellant-case.test.js
zoebutterworth: The new rows will be added as a separate ticket. The only change in comparison to HAS in this work is that we no longer have the 'Additional documents' section, which is why I check that the page does not contain that

PR #1561, opened 2025-06-23
Feat/s20 seed data

andy-mair121:
## Describe your changes

Add S20 lpaq seed data and a range of cases at different states

### DIFF
--- a/appeals/api/src/database/seed/data-test.js
+++ b/appeals/api/src/database/seed/data-test.js
@@ -254,12 +254,6 @@ const newS78Appeals = [
 		assignCaseOfficer: false,
 		agent: false
 	}),
-	appealFactory({ typeShorthand: APPEAL_CASE_TYPE.Y, assignCaseOfficer: false }),
-	appealFactory({
-		typeShorthand: APPEAL_CASE_TYPE.Y,
-		siteAddressList: addressListForTrainers,
-		assignCaseOfficer: false
-	}),
 	appealFactory({
 		typeShorthand: APPEAL_CASE_TYPE.W,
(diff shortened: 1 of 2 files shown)


appeals/api/src/database/seed/data-test.js
```diff
@@ -515,6 +509,49 @@ const newS78Appeals = [
...
+		assignCaseOfficer: false
+	}),
+	appealFactory({
+		typeShorthand: APPEAL_CASE_TYPE.Y,
+		status: { status: APPEAL_CASE_STATUS.VALIDATION, createdAt: getPastDate({ days: 5 }) },
+		lpaQuestionnaire: true,
+		assignCaseOfficer: true,
+		agent: false
+	}),
+	appealFactory({
```
toyin-pins, line 526: We should have a case that's in lpaq status but NOT received the lpaq yet.

appeals/api/src/database/seed/data-test.js
```diff
@@ -515,6 +509,49 @@ const newS78Appeals = [
...
+		typeShorthand: APPEAL_CASE_TYPE.Y,
+		status: { status: APPEAL_CASE_STATUS.FINAL_COMMENTS, createdAt: getPastDate({ months: 10 }) },
+		lpaQuestionnaire: true,
+		startedAt: getPastDate({ months: 6 }),
+		validAt: getPastDate({ months: 10 }),
+		assignCaseOfficer: false,
+		agent: true
```
toyin-pins, line 551: think we could also have a couple more cases with agents

toyin-pins (changes requested):

PR #1594, opened 2025-06-27
fix(appeals): fixed issue with status of existing hearing after state…

peter-uzor-pins:
…ment

## Describe your changes

fixed issue with status of existing hearing appeal after assuming hearing is entered before statement is completed

[A2-3743](https://pins-ds.atlassian.net/browse/A2-3743)

[A2-3743]: https://pins-ds.atlassian.net/browse/A2-3743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/state/transition-state.js
+++ b/appeals/api/src/server/state/transition-state.js
@@ -73,11 +73,12 @@ const transitionState = async (appealId, azureAdUserId, trigger) => {
 	});
 
 	if (
-		currentState !== APPEAL_CASE_STATUS.AWAITING_EVENT &&
 		newState === APPEAL_CASE_STATUS.EVENT &&
-		// newState === APPEAL_CASE_STATUS.ISSUE_DETERMINATION &&
 		[APPEAL_TYPE_SHORTHAND_HAS, APPEAL_TYPE_SHORTHAND_FPA].includes(appealType.key) &&
-		appeal.siteVisit
+		((appeal.procedureType?.key === APPEAL_CASE_PROCEDURE.WRITTEN && appeal.siteVisit) ||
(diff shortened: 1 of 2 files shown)


appeals/api/src/server/state/transition-state.js
```diff
@@ -73,11 +73,12 @@ const transitionState = async (appealId, azureAdUserId, trigger) => {
...
 	if (
-		currentState !== APPEAL_CASE_STATUS.AWAITING_EVENT &&
 		newState === APPEAL_CASE_STATUS.EVENT &&
-		// newState === APPEAL_CASE_STATUS.ISSUE_DETERMINATION &&
 		[APPEAL_TYPE_SHORTHAND_HAS, APPEAL_TYPE_SHORTHAND_FPA].includes(appealType.key) &&
-		appeal.siteVisit
+		((appeal.procedureType?.key === APPEAL_CASE_PROCEDURE.WRITTEN && appeal.siteVisit) ||
+			(appeal.procedureType?.key === APPEAL_CASE_PROCEDURE.HEARING &&
+				appeal.hearing &&
+				appeal.hearing?.addressId))
```
jimbali, line 81: lgtm but is there a test that we can write for it?

peter-uzor-pins: Oh yes, i forgot to do that. Will add some tests.

peter-uzor-pins: Actually looking at it again. It's going to be quite difficult writing tests for this section as this area is very integrated and the dependency on mocking is going to be massive meaning it wouldn't be a true reflection.
So for now probably will just leave the tests. 

I have tested manually and it was working fine.

PR #1596, opened 2025-06-27
chore(tooling): run format-check on PR

beejjacobs:
https://pins-ds.atlassian.net/browse/RDSD-138

Setup a format-check to ensure Prettier config is being adhered to. With pre-commit hooks its possible to skip so needs to be checked in a pipeline.


ypk (dismissed):
Looks good to me 👍🏻

PR #1611, opened 2025-07-01
feat(web): field highlighting and messages for date errors

lachlandow1:
### WEB
1. Updated date fields to highlight only the correct empty fields for standard dates and document dates
2. Refactored both validators so that the checks follow the same format
3. Updated css class mappers to accommodate new setup and be more easily adapted if more changes come in the future

### TESTS

1. Updated Unit tests to accomodate for new setup
2. Updated snapshots
3. Manually checked date fields

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-1741)


toyin-pins (approved):
LGTM

PR #1631, opened 2025-07-04
test(api): a2-3575 s20 appellant notify p1

andy-mair121:
## Describe your changes

Adds missed appellant notify test cases

[Ticket A2-3575](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3575)


appeals/api/src/server/endpoints/appellant-cases/__tests__/appellant-cases.test.js
zoebutterworth, line 1: Could you add the generalisation, as has been added in other places, to the notify on approx lines 579-590, 859-857, 932-940?

PR #1649, opened 2025-07-08
Feature/a2 3234 error messages and success banner in related appeals

praveenmurugan743:

[Ticket]()


praveenmurugan743:
> Looks good overall, may I suggest that we please add 2 extra tests to cover the new validation? Appeal reference contains mixed letters/numerics e.g. 12345AB and ref all letters e.g. ABCDEFG

Thank you, I have added two unit tests to cover this validation.

PR #1659, opened 2025-07-09
fix(A2-3846): converting dates to utc datetime in web before sending api request

toyin-pins:
## Describe your changes
- Fixing bug where date changes caused API to crash with 400 bad requests. Now formatting the date on the web side of things before sending request to API. 
- Declared missing modules which caused issue for business day check
- Added logging for error

[Ticket](https://pins-ds.atlassian.net/browse/A2-3846)


jimbali:
Needs tests for sure, but if you need to fix it quickly you could do it in another PR I guess, or if you're doing a proper fix immediately afterwards then I guess it will change anyway...

PR #1687, opened 2025-07-16
fix(appeals): use correct keys for developmentType

jimbali:
## Describe your changes

- Move APPEAL_DEVELOPMENT_TYPES constant to shared package to keep it consistent between web & API
- Update APPEAL_DEVELOPMENT_TYPES keys to match the data coming from FO

My only concern here is that if somebody has edited the values in the Back Office then the wrong keys may have been stored in the DB, so we will need to check that and update any wrong values.

[A2-3902](https://pins-ds.atlassian.net/browse/A2-3902)

[A2-3902]: https://pins-ds.atlassian.net/browse/A2-3902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


toyin-pins (approved):
LGTM

PR #1691, opened 2025-07-17
chore(appeals): changing pins-data-model to @planning-inspectorate/da…

zoebutterworth:
…ta-model (A2-3900)

## Describe your changes

Updating to data-model version 1.11.0 and moving to using the npm version instead of github

Part of [Ticket](https://pins-ds.atlassian.net/browse/A2-3900)


beejjacobs (changes requested):
Looks good, only small change is to update the root package.json which still includes the github reference.

PR #1715, opened 2025-07-22
Test/a2 3826 linked appeals

jwillis-pins:
## Describe your changes

# Linked Appeals

* Refactored Linked Appeals test cases
* Added `.skip` on unlinked cases whilst they undergo dev work
* In anticipation of future dev work, added missing negative test cases for Linked Appeals (1 skipped)
* Refactored `cy.checkNotifySent()` function to now accept a single or multiple objects. These objects should could a template and recipient properties. This was to fix the issue where multiple of the same template were sent but to different recipients. We can now assert on these unique combinations
* Fixed test cases that were using the previous version of `cy.checkNotifySent()`
* Fixed some small issues in the hearing.spec whilst I was there

[a2 3826](https://pins-ds.atlassian.net/browse/A2-3826)


RHARV018 (dismissed):
Looks good to me!

PR #1722, opened 2025-07-23
fix(appeals): make changeAppealType optional for db migration

matt2415:
## Describe your changes

- Db migration to make `changeAppealType` optional

[Ticket](https://pins-ds.atlassian.net/browse/A2-3615)


jimbali (changes requested):
Unapproving because on further discussion we should probably edit the previous migration instead as it is broken

PR #1758, opened 2025-07-30
fix(A2-3330): adding total uploaded file size client side validation

toyin-pins:
## Describe your changes
- Added client side validation for the total file size of uploaded files 
- Files are not added if new uploads would take total beyond 1GB 
- Updated the error message copy
- Added constants for clarity of byte conversions

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3330)


BenSurgisonPINS (approved):
lgtm

PR #1761, opened 2025-07-30
fix(api): changing site visit and costs to display for procedure type…

zoebutterworth:
… unknown (A2-3982)

## Describe your changes

- Updating the site visit and costs to display when the procedure type is undefined (we treat undefined in the same way as written)
- Adding tests for the procedure type undefined case

[Ticket](https://pins-ds.atlassian.net/browse/A2-3982)


toyin-pins (approved):
lgtm

PR #1771, opened 2025-07-31
fix(api): extend file upload hex signature list to allow common uploads

bryn-pins:
## Describe your changes

Extending the list of valid hex signatures used when validating mime types
Added a few more from https://en.wikipedia.org/wiki/List_of_file_signatures

[Ticket](https://pins-ds.atlassian.net/browse/A2-4039)


andy-mair121 (approved):
lgtm

PR #1787, opened 2025-08-05
Added e2e tests for starting case as procedure type - inquiry

jwillis-pins:
## Describe your changes

Added e2e tests for starting case as procedure type - inquiry

[A2-3499](https://pins-ds.atlassian.net/browse/A2-3499)


RHARV018 (approved):
This looks like there's everything there now - thank you very much!

PR #1809, opened 2025-08-07
Test/a2 3950 adding related appeals in lpaq and appellant case

jwillis-pins:
## Describe your changes

* Added test cases for adding related appeals from the appellant case and lpaq pages
* Fixed banner message assertions on existing related appeals test cases

[a2-3950](https://pins-ds.atlassian.net/browse/A2-3950)


RHARV018 (dismissed):
Looks all good to me!

PR #1814, opened 2025-08-08
feat(appeals): hiding decision row on overview until issue decision

matt2415:
## Describe your changes

Hiding decision row on overview section until issue decision stage

[Ticket](https://pins-ds.atlassian.net/browse/A2-3952)

### DIFF
--- a/appeals/web/src/server/lib/mappers/data/appeal/submappers/decision.mapper.js
+++ b/appeals/web/src/server/lib/mappers/data/appeal/submappers/decision.mapper.js
@@ -16,6 +16,10 @@ import config from '#environment/config.js';
 export const mapDecision = ({ appealDetails, session, request }) => {
 	const { appealId, appealStatus, decision } = appealDetails;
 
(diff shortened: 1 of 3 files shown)


appeals/web/src/server/lib/mappers/data/appeal/submappers/decision.mapper.js
```diff
@@ -16,6 +16,10 @@ import config from '#environment/config.js';
 export const mapDecision = ({ appealDetails, session, request }) => {
 	const { appealId, appealStatus, decision } = appealDetails;
 
+	if (!isStatePassed(appealDetails, APPEAL_CASE_STATUS.AWAITING_EVENT)) {
+		return { id: 'decision', display: {} };
+	}
```
jimbali, line 21: There could be an issue with this - I think it can sometimes go from `awaiting_event` back to `event`, for example when the address is removed from a hearing, but in that case the `isStatePassed` function I think would still return true for `awaiting_event`.

PR #1830, opened 2025-08-11
test/A2-3259 Adding tests for date error handling 

marcpodgorney81:
## Describe your changes 

In this PR am adding a set of tests for the date validations as outlined in this document https://pins-ds.atlassian.net.mcas.ms/wiki/spaces/BO/pages/2314698826/Global+Date+Validation 

Also as part of this have introduced a new `InquirySectionPage` as seemed a logical and clean approach - is very basic atm but going forward could potentially move some logic away from the monolith `CaseDetailsPage` into it e.g.?  

In addition did some small refactoring to move `verifyErrorMessages` into the base page - as realised is now shared by hearing and enquiry pages, also it's dependent functions are in base page as well so seemed to make sense? 

*fyi time validation is not yet working (ticket raised for this https://pins-ds.atlassian.net.mcas.ms/browse/A2-4100?McasTsid=20596&McasCtx=4) - when implemented tests for this will be in follow up PR* 

**Question - should these tests be run for hearing also - or given that is global validation should this be enough to give confidence that is working?** 

https://pins-ds.atlassian.net.mcas.ms/browse/A2-3259


appeals/e2e/cypress/page_objects/basePage.js
```diff
@@ -371,6 +371,19 @@ export class Page {
 		cy.focused().should('have.attr', attribute, value);
 	}
 
+	verifyErrorMessages(options) {
```
marcpodgorney81, line 374: Moving here from hearing and inquiry pages as seems logical? (e.g. `verifyInputFieldIsFocusedWhenErrorMessageLinkIsClicked` and `verifyInlineErrorMessage` are both in this class)

hardiebrown1:
Hi @marcpodgorney81 this was just for dateValidation.spec.js. 

Spoke with one of the devs and they said they have extensive units tests for the date validation, around 100. 

It's a global component so we could say this test acts as coverage for all date inputs across the site.

marcpodgorney81:
Hey @hardiebrown1 @jwillis-pins 

Thanks for all the great feedback! 

Few answers in response to your questions/comments 

- @jwillis-pins Yes these tests are for global validation - inquiry is the example as that is the ticket I was working on - think will rename the file to `globalDateValidation` e.g. so that purpose is clearer  

- So in regard the tests taking 10 minutes - am surprised at that, as each time I've run them its taken around 2.5 minutes (3 max) - was that consistent for you? 

- And am happy to move forward with @jwillis-pins suggestion! - was also looking at if could be more efficient, and this is one way!  

Thanks!

hardiebrown1:
Hi @marcpodgorney81 run the tests against dev or test. 

Ran them against the dev environment. The pipeline runs against dev.
image

marcpodgorney81:
Ok, will try that @hardiebrown1

hardiebrown1:
@marcpodgorney81 @jwillis-pins pushed some code that removes test isolation for this spec. Got the runtime down to 1 minute 15 seconds

hardiebrown1 (approved):
LGTM

jwillis-pins (approved):
Looking good!

PR #1832, opened 2025-08-11
feat(api): updating state machine to allow transition to invalid (A2-4010)

lachlandow1:
### API

- Added state transitions to invalid for required states

### TEST 

- Added unit test for these transitions

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-4010)

### DIFF
--- a/appeals/api/src/server/state/create-state-machine.js
+++ b/appeals/api/src/server/state/create-state-machine.js
@@ -64,7 +64,8 @@ const createStateMachine = (appealType, procedureType, currentState, eventElapse
 					},
 					[APPEAL_CASE_STATUS.WITHDRAWN]: {
 						target: APPEAL_CASE_STATUS.WITHDRAWN
-					}
+					},
+					[APPEAL_CASE_STATUS.INVALID]: { target: APPEAL_CASE_STATUS.INVALID }
 				},
 				meta: {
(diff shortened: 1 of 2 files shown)


appeals/api/src/server/state/create-state-machine.js
```diff
@@ -86,7 +86,8 @@ const createStateMachine = (appealType, procedureType, currentState, eventElapse
 					},
 					[APPEAL_CASE_STATUS.WITHDRAWN]: {
 						target: APPEAL_CASE_STATUS.WITHDRAWN
-					}
+					},
+					[APPEAL_CASE_STATUS.INVALID]: { target: APPEAL_CASE_STATUS.INVALID }
```
toyin-pins, line 90: Need to add this line to the above block also for VALIDATION. Currently can only make a case invalid during validation in validate appeal flow. But adding block will also allow making it invalid through cancel appeal flow

toyin-pins (changes requested):

appeals/api/src/server/state/create-state-machine.js
lachlandow1: state transition for validation stage added

toyin-pins (approved):

PR #1858, opened 2025-08-15
fix(appeals): refactor notify for on linking appeals

matt2415:
## Describe your changes

* Updating copy for notify sent on linking appeals
* New function to determine which copy should be sent
* Test updates and new coverage

[Ticket](https://pins-ds.atlassian.net/browse/A2-4002)


BenSurgisonPINS (approved):
lgtm

PR #1913, opened 2025-08-28
feat(api): a2-3858 cas adverts appeal form

andy-mair121:
## Describe your changes

Api work to enable additional rows required for cas adverts appeals

[Ticket](https://pins-ds.atlassian.net/browse/A2-3858)

### DIFF
--- a/appeals/api/src/server/mappers/api/definitions/appellant-case.js
+++ b/appeals/api/src/server/mappers/api/definitions/appellant-case.js
@@ -1,10 +1,10 @@
-import { InvalidIncompleteReason } from './invalid-incomplete.js';
-import { Folder } from './folders-documents.js';
-import { AppealSummary } from './appeal-summary.js';
 import {
 	APPEAL_DEVELOPMENT_TYPE,
 	APPEAL_TYPE_OF_PLANNING_APPLICATION
 } from '@planning-inspectorate/data-model';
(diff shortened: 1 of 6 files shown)


appeals/api/src/server/mappers/api/definitions/appellant-case.js
```diff
@@ -338,6 +338,26 @@ export const AppellantCaseUpdateRequest = {
...
+			nullable: true
+		},
+		advertInPosition: {
+			type: 'number',
+			nullable: true
+		},
+		landownerPermission: {
+			type: 'number',
+			nullable: true
+		},
```
zoebutterworth, line 353: Should these be 'string'? From looking above it looks like booleans are of type string

appeals/api/src/server/openapi-types.ts
```diff
@@ -12192,6 +12192,11 @@ export interface AppellantCaseUpdateRequest {
 	originalDevelopmentDescription?: string | null;
 	changedDevelopmentDescription?: string | null;
 	numberOfResidencesNetChange?: number | null;
+	highwayLand?: number | null;
+	advertInPosition?: number | null;
+	landownerPermission?: number | null;
```
zoebutterworth, line 12197: Similarly here- should they be string?

PR #1914, opened 2025-08-28
A2-3705 Refactoring Issue Decision spec file

jwillis-pins:
## Describe your changes

* Added cases for issuing a decision with partial costs - appellant only, lpa only
* Modified happyPathHelpers to allow for finer control over which costs to also issue decisions for. These are controlled via a bool in the arguments
* Removed redundant `issueDecisionWithoutCosts` function. Replaced usage with `issueDecision` with additional args instead
* Consolidated 3 spec files relating to issuing decisions into 1 file

[A2-3705](https://pins-ds.atlassian.net/browse/A2-3705)

[A2-3705]: https://pins-ds.atlassian.net/browse/A2-3705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RHARV018 (approved):
Looks good to me!

PR #1921, opened 2025-08-29
fix(api): broadcast updated details for both child and lead linked appeals

topickering:
When two appeals are linked, this change ensures that the updated details in respect of both linked appeals are broadcast.

[Ticket]https://pins-ds.atlassian.net/browse/A2-3911


BenSurgisonPINS (approved):
lgtm

PR #1936, opened 2025-09-01
fix(appeals): use lead appeal when importing representations (a2-3690)

BenSurgisonPINS:
#### Make sure imported representations use the lead appeal if they are a for a linked child appeal (a2-3690)

This is a bug fix for a previous ticket
 
### API:
- When validating and loading the appeal in the integrations middleware for representations, make sure the representations are loaded on to the lead appeal if the appeal reference supplied is for a linked child appeal.
- Remove unused imports in database schema 

### Test:
- Add extra unit tests
- Make sure tests still pass
- Tested with swagger
- Test within browser

[A2-3690 - Reviewing interested party (IP) comments for linked written rep S78 planning appeals](https://pins-ds.atlassian.net/browse/A2-3690)

### DIFF
--- a/appeals/api/src/server/endpoints/integrations/integrations.middleware.js
+++ b/appeals/api/src/server/endpoints/integrations/integrations.middleware.js
@@ -6,10 +6,13 @@ import {
 	ERROR_INVALID_APPELLANT_CASE_DATA,
 	ERROR_INVALID_LPAQ_DATA,
 	ERROR_INVALID_REP_DATA,
-	ERROR_INVALID_APPEAL_TYPE_REP
+	ERROR_INVALID_APPEAL_TYPE_REP,
+	CASE_RELATIONSHIP_LINKED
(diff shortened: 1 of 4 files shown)


appeals/api/src/server/endpoints/integrations/integrations.middleware.js
```diff
@@ -122,7 +125,8 @@ export const validateRepresentation = async (req, res, next) => {
 		});
 	}
 
-	const referenceData = await loadReferenceData(body?.caseReference);
+	const useLeadAppealIfLinked = featureFlags.isFeatureActive(FEATURE_FLAG_NAMES.LINKED_APPEALS);
```
toyin-pins, line 128: Shouldn't this also be checking if the appeal itself is actually the child appeal? Because currently if the appeal IS the parent, we'll be executing a bunch of code we don't need to?

PR #1939, opened 2025-09-02
Feat/a2 3790 update national filters list

lachlandow1:
### API
- added new seed function to assign teams to appeals according to lpa assignments - included some null mappings to have unassigned appeals when seeding
- updated production see data to match ticket 
- updated dev seed data so that different scenarios could be tested locally  
- updated appeals endpoint to filter or assignedTeamId
- updated unit tests to cover filtering  on assignedTeamId

### WEB

- added casse team filter to national list page
- added new unite tests and snapshots for national list to cover changes 
- updated submapper and CYA page to cover logic for null emails of teams
- updated unit test and snapshots  
- added new tests for filtering on assignedTeamId 
- added tests to cover scenarios for CYA and appeal details page

[Ticket](https://pins-ds.atlassian.net.mcas.ms/browse/A2-3721)

### DIFF
--- a/appeals/api/src/database/seed/add-assigned-team-id.js
+++ b/appeals/api/src/database/seed/add-assigned-team-id.js
@@ -1,0 +1,13 @@
+/**
+ * @param {import('#db-client').PrismaClient} databaseConnector
+ */
+export const addAssignedTeamIdToAppeal = async (databaseConnector) => {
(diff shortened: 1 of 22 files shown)


toyin-pins (commented):
LGMT - just need to squash commits

PR #1950, opened 2025-09-03
fix(appeals): set up site visit banner showing correctly for HAS and …

zoebutterworth:
…CAS planning (A2-4274)

## Describe your changes

Using procedureType rather than the one associated to the appeal so that there is no undefined case because we have defaulted it to written

Removed a rogue console.log

Tested manually

Created a tech debt ticket to add testing to the transitionState function: https://pins-ds.atlassian.net/browse/A2-4315

[Ticket](https://pins-ds.atlassian.net/browse/A2-4274)


BenSurgisonPINS (approved):
lgtm

lachlandow1 (approved):

PR #1969, opened 2025-09-05
feat(web): remove case details accordion (A2-2107)

lachlandow1:
### WEB

- removed the accordion component usage from appeal details screen - replaced headings in common mappers 
- Updated unit tests and snapshots  
  
[Ticket](https://pins-ds.atlassian.net/browse/A2-2107)


toyin-pins (approved):
lgtm

PR #1988, opened 2025-09-09
feat(api): update training seed data to include production data (A2-4373)

lachlandow1:
### API

1. update training seed script to add the production teams and team assignments
2. updated the training seed data to use an ID that will never be used in the sequence
3. tested locally

[Ticket](https://pins-ds.atlassian.net/browse/A2-4373)


praveenmurugan743 (approved):
LGTM!

PR #1992, opened 2025-09-09
Feature/a2 2107 removing case details accordion and fix e2e tests

lachlandow1:
### WEB

- removed the accordion component usage from appeal details screen - replaced headings in common mappers
- added feature flag  
- Updated unit tests and snapshots

[Ticket](https://pins-ds.atlassian.net/browse/A2-2107)


BenSurgisonPINS (approved):
lgtm

PR #2051, opened 2025-09-16
fix(web): Validating IP params before calling api

toyin-pins:
## Describe your changes
- This is an attempt to fix consistent Server-side request forgery issues flagged by Github code scans. 
- The idea is that ID parameters used in API calls will be validated just before making the API call. Especially in cases where these IP params are taken from the query parameters. 

Hopefully this will resolve the top 2 scan issues: 
1. appeals/web/src/server/appeals/appeal-details/lpa-questionnaire/changed-listed-buildings/changed-listed-buildings.controller.js
2. appeals/web/src/server/appeals/appeal-details/representations/interested-party-comments/interested-party-comments.middleware.js
Screenshot 2025-09-17 at 14 41 05


andy-mair121 (commented):
Lgtm

andy-mair121 (approved):
Lgtm

PR #2060, opened 2025-09-17
feat(appeals): make seed data for reasons updatable by id

ShareMyWebStuff:
## Describe your changes

Currently if you change the seed data name it will create another reason. This change with remove the autoincrement from the tables so we control the id column, the update is also done on the id so the name can be changed.

**Note**

Currently the training environment has a different id to the rest of the environments. We will need to either

1. truncate the table to reset the autoincrement (we may not have permissions as truncate doesn't log)
2. rerun the seed data

or
1. Run the seed
2. Manipulate the id's with SQL.


ShareMyWebStuff:
Not required

PR #2073, opened 2025-09-18
A2 3898/e2e parallelization

hardiebrown1:
## Describe your changes

To improve CI efficiency and reduce overall test execution time, we are implementing parallelization of Cypress E2E tests using Azure Pipelines' strategy.matrix feature.

There is three main changes that have been made in order to reduce overall execution time:

- YML changes in order to run the tests across 3 parallel agents.
 Running the tests with three agents we had reduced the execution from 1 hour and 5 minutes to 30 minutes.
image- 

- Update MS authentication to use cy.session. 
Cy.session is widely used by industry in order to store cookies. The benefit being the cookies will persist across the whole session. Currently the azure sign in method is called multiple times throughout a pipeline run.

- Test go directly to the case details page & bypass /all-cases.
API calls to the all-cases page was causing the most stress on the system leading to timeout failures. I have updated the createCase method to bring in the appeal ID and the appeal Reference. Create Case now returns and object as opposed to a string. 

In order to minimise code changes I have updated the helper methods to accept an object instead of passing in a string. However some methods on individual case pages still accept strings.

I have kept the object with name - "caseRef". Should this be changed to "caseObj"? 

[A2-3898](https://pins-ds.atlassian.net/browse/A2-3898)


marcpodgorney81:
Hey @hardiebrown1, great PR! 👍  
Should bring some great improvement to test run time! 

Some comments/suggestions that have so far 

- for the object containing id/reference, was wondering about `caseDetails` or `appealDetails` as a name?  
  - logical name/grouping  
  - e.g. `caseDetails.reference` or `appealDetails.id` 
  
- would suggest to share this with developers also, as will be especially interested in the change to how runs in pipeline! 
  - maybe a short session to go through changes to cypress config/commands?

hardiebrown1:
> Hey @hardiebrown1, great PR! 👍 Should bring some great improvement to test run time! 💯
> 
> Some comments/suggestions that have so far
> 
> * could these changes affect how we run tests locally, e.g. using `npm run cy:open`? (doesn't seem so, just checking!)
Won't change anything locally. Cy. session and directing straight to case details should reduce execution time locally however.
> * for the object containing id/reference, was wondering about `caseDetails` or `appealDetails` as a name?
>   * logical name/grouping
>   * e.g. `caseDetails.reference` or `appealDetails.id`
>   
>   So then e.g.
>   ```
>   cy.createCase({ caseType: 'W' }).then((caseDetails) => {
>   }
>   cy.createCase({ caseType: 'W' }).then((appealDetails) => {
>   }
>   ```
Rename to caseObj.
> * your change to `createCase` command in `commands.js` - like the update to logging! just wondering why re-defining the object, and if could have been written like this e.g.?
>   ```
>   return cy.wrap(null).then(async () => {
>     const caseDetails = await appealsApiClient.caseSubmission(customValues);
>     cy.log(`Generated case with ref ${caseDetails.reference} and id ${caseDetails.id}`);
>     return caseDetails; 
>   });
>   ```
>   I don't want to break anything if it's working as is. Could implement this in a future ticket.
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   is fine as is though!
> * with regards to reporting, think format is ok - maybe update naming to something like `Run appeals-backoffice-tests group{number}` e.g.
> 
> ```
> Run appeals-backoffice-tests group1
> Run appeals-backoffice-tests group2
> Run appeals-backoffice-tests group3 
> ```
> 
> may bit bit clearer as to which app it is
I will add this to the reporting ticket and get it fixed there.
> * would suggest to share this with developers also, as will be especially interested in the change to how runs in pipeline!
>   
>   * maybe a short session to go through changes to cypress config/commands, and how will work with nightly run e.g.?

appeals/e2e/cypress/e2e/back-office-appeals/changeAppealProcedureType.spec.js
```diff
@@ -185,11 +185,11 @@ describe('change appeal procedure types', () => {
 	const setupTestCase = () => {
 		cy.login(users.appeals.caseAdmin);
 		cy.createCase({ caseType: 'W' }).then((ref) => {
```
marcpodgorney81, line 187: Same as above!

appeals/e2e/cypress/page_objects/caseDetailsPage.js
```diff
@@ -470,13 +470,13 @@ export class CaseDetailsPage extends Page {
 		this.elements.costDecisionStatus().contains(text);
 	}
 
-	clickRemoveRelatedAppealByRef(caseRefToRelate) {
-		cy.log(caseRefToRelate);
-		cy.getByData('remove-appeal-' + caseRefToRelate).click();
+	clickRemoveRelatedAppealByRef(caseObjToRelate) {
+		cy.log(caseObjToRelate);
+		cy.getByData('remove-appeal-' + caseObjToRelate).click();
```
marcpodgorney81, line 475: does this need a `.reference` or `.id`? e.g. `cy.getByData('remove-appeal-' + caseObjToRelate.reference)`, or does it auto toString or something?

appeals/e2e/cypress/page_objects/listCasesPage.js
```diff
@@ -40,7 +40,7 @@ export class ListCasesPage extends Page {
 	}
 
 	clickAppealByRef(ref) {
```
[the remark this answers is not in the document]
hardiebrown1: yes that would be better

will also add this to improvements ticket

PR #2084, opened 2025-09-19
HotFix: Release rollback state causing deletions

toyin-pins:
## Describe your changes
The rollback state was not scoped to a particular appeal which caused many appealStatus deletions to occur after some SQL queries had been applied

[Ticket]()


BenSurgisonPINS (approved):
lgtm

PR #2161, opened 2025-09-30
fix(appeals): show select case management links on awaiting transfer

matt2415:
## Describe your changes

Enabling Documentation and Case Management links for appeals in Awaiting Transfer

[Ticket](https://pins-ds.atlassian.net/browse/A2-3232)


BenSurgisonPINS (approved):
lgtm

PR #2198, opened 2025-10-06
fix(test/e2e): fix missing variable

abitec:
## Describe your changes

- Fixed variable reference from ref to caseRef in the S78 hearing case test.
- Once Bug A2-4636 is resolved, standard date and time format should be used as noted in the added comment.

[Ticket]()


marcpodgorney81 (approved):
LGTM!

PR #2724, opened 2025-12-11
test(appeals): updating api command names and creating new command files

marcpodgorney81:
## Describe your changes

In this PR am **making some changes to the cypress commands setup** 

**Renaming api commands** 
Have renamed the api commands to remove the `viaApi` from some names, as was causing some confusion, and also some inconsistency! 

**New command files**
The existing `commands.js` file was getting quite large and bit difficult to follow, as different types of command were mixed up, i.e. trying to keep track of what api commands we have 
as such have moved them into separate files, for clarity and to make it easier to find command that are looking for! 

- commands/api.js - commands relating to api client (hearing, inquiry etc.) 
- commands/authentication.js - commands relating to authentication, login etc. 
- commands/utils.js - helper commands (e.g. find elements) 

These are now each imported into the e2e.js file            

https://pins-ds.atlassian.net/browse/A2-7731


marcpodgorney81:
Am going to close this as are large number of conflicts, think will be easier/cleaner to start new branch with the changes

PR #3348, opened 2026-03-03
feat: adding private endpoint for key vault

m-juckes-pins:
## Describe your changes

This PR will create an endpoint for the key vault so the secrets are not public.

[Ticket]()


DavidJWhiteDev (dismissed):
lgtm

DavidJWhiteDev (approved):
chatted with Juckes, all good

PR #3478, opened 2026-03-18
chore(appeals): performance todos and deprecations

bryn-pins:
## Describe your changes

This PR attempts to stop new uses of the inefficient calls that return all data for an appeal
- renames some deprecated functions so that they are more obvious to spot in PR review
- patch appeal-details was using ?include=all which is deprecated and unnecessary, the response isn't used, so the endpoint itself can decide on which lookups are needed
- added comments for some performance issues, tickets exist as well
- encoded uri based on github suggestion - unrelated to my changes so left in a separate commit

[Ticket](https://pins-ds.atlassian.net/browse/A2-7612)

### DIFF
--- a/appeals/api/src/server/endpoints/appeal-details/__tests__/appeals-details.test.js
+++ b/appeals/api/src/server/endpoints/appeal-details/__tests__/appeals-details.test.js
@@ -306,7 +306,7 @@ describe('Appeal detail routes', () => {
 				});
 
 				const response = await request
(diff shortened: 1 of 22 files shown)


DavidJWhiteDev (approved):
very clear, nice!

DavidJWhiteDev (approved):
all runs good, and clearly shows where we need to make improvements - and clear to other devs not to use, nice one!

PR #3591, opened 2026-04-01
fix(web): fixed incorrect banner showing when USV selected (A2-7805)

musabadam-pins:
## Describe your changes

- Updated banner logic so it appears based on siteVisitType as well as the time and date values

[Ticket](https://pins-ds.atlassian.net/browse/A2-7805)

### DIFF
--- a/appeals/web/src/server/lib/mappers/utils/__tests__/map-status-dependent-notifications.test.js
+++ b/appeals/web/src/server/lib/mappers/utils/__tests__/map-status-dependent-notifications.test.js
@@ -2,9 +2,13 @@
 import { mapStatusDependentNotifications } from '#lib/mappers/utils/map-status-dependent-notifications.js';
 import { appealDataToGetRequiredActions } from '#testing/appeals/appeals.js';
 import { APPEAL_TYPE } from '@pins/appeals/constants/common.js';
-import { VALIDATION_OUTCOME_INCOMPLETE } from '@pins/appeals/constants/support.js';
+import {
+	SITE_VISIT_TYPE_ACCESS_REQUIRED,
+	SITE_VISIT_TYPE_ACCOMPANIED,
+	SITE_VISIT_TYPE_UNACCOMPANIED,
+	VALIDATION_OUTCOME_INCOMPLETE
+} from '@pins/appeals/constants/support.js';
 import { APPEAL_CASE_STATUS } from '@planning-inspectorate/data-model';
-
 describe('mapStatusDependentNotifications', () => {
 	const mockAppealData = {
 		appealId: 1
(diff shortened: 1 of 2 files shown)


appeals/web/src/server/lib/mappers/utils/required-actions.js
```diff
@@ -511,10 +511,31 @@ export function getRequiredActionsForAppeal(appealDetails, view) {
 
 	// @ts-ignore
 	const siteVisit = appealDetails?.siteVisit;
-	if (siteVisit && !siteVisit.visitDate && !siteVisit.visitStartTime) {
-		actions.push('addSiteVisitDateTime');
-	} else if (siteVisit && siteVisit.visitDate && !siteVisit.visitStartTime) {
-		actions.push('addSiteVisitTime');
+	if (siteVisit && siteVisit.visitType) {
+		switch (siteVisit.visitType.toLowerCase()) {
```
zoebutterworth: I feel that this switch is a bit unnecessary. There is quite a lot of repetition in each of the cases. I think you should be able to just add an extra statement to the if to check for unaccompanied

zoebutterworth (commented):
Are there some tests which cover this area of the codebase which you could add to to cover this case?

appeals/web/src/server/lib/mappers/utils/required-actions.js
```diff
@@ -513,10 +515,18 @@ export function getRequiredActionsForAppeal(appealDetails, view) {
 
 	// @ts-ignore
 	const siteVisit = appealDetails?.siteVisit;
-	if (siteVisit && !siteVisit.visitDate && !siteVisit.visitStartTime) {
-		actions.push('addSiteVisitDateTime');
-	} else if (siteVisit && siteVisit.visitDate && !siteVisit.visitStartTime) {
-		actions.push('addSiteVisitTime');
+	if (siteVisit && siteVisit.visitType) {
+		const visitType = siteVisit.visitType;
+		if (!siteVisit.visitDate) {
```
zoebutterworth, line 520: Did you mean to remove the additional check for !siteVisit.visitStartTime here?

appeals/web/src/server/lib/mappers/utils/__tests__/map-status-dependent-notifications.test.js
zoebutterworth, line 311: Ideally there would be a test that addSiteVisitTime won't show up for unaccompanied appeals

appeals/web/src/server/lib/mappers/utils/required-actions.js
musabadam-pins: its been a long time but i believe so, at the minimum you would have the date before having a start time, the next set of conditions check if startTime exists depeneding on the type of siteVisit

appeals/web/src/server/lib/mappers/utils/__tests__/map-status-dependent-notifications.test.js
```diff
@@ -320,12 +325,54 @@ describe('mapStatusDependentNotifications', () => {
 			requiredAction: 'addSiteVisitDateTime',
 			mockData: {
 				appealId: 1,
-				appealStatus: APPEAL_CASE_STATUS.INVALID,
-				siteVisit: {}
```
zoebutterworth, line 324: Did you mean to remove this test for APPEAL_CASE_STATUS.INVALID?

musabadam-pins: yep, it was testing a scenario that was not possible and was failing

zoebutterworth: Following further discussion I am happy with this test no longer being relevant

zoebutterworth (approved):

PR #3605, opened 2026-04-08
fix(api): fix validation errors message (A2-7611)

DavidJWhiteDev:
## Describe your changes

Small fix to correct the message logged with there is an express middleware validation error.  Prior to this it simply wrote "Validation errors" in the log, and did not write out the actual error.

## A2-7611 BO Tech Debt: Error message just says Validation errors and not the actual errors

[Ticket](https://pins-ds.atlassian.net/browse/A2-7611)


julietteorpen (approved):
LGTM!

PR #3606, opened 2026-04-08
feat(api): don't broadcast submission docs until after scan (A2-7132)

bryn-pins:
## Describe your changes

- don't broadcast documents from external submissions, wait for virus scan
- manually uploaded docs in BO will more than likely virus scan before upload as it's done client side, broadcast will happen after check your answers instead
- if a manual upload is saved before the virus scan is complete, the virus scan will rebroadcast

[Ticket](https://pins-ds.atlassian.net/browse/A2-7132)


DavidJWhiteDev (approved):
makes sense!

PR #3613, opened 2026-04-14
feat(web): add google tag manager (A2-7537)

topickering:
## Describe your changes

Add google tag to head component

[Ticket](https://pins-ds.atlassian.net/browse/A2-7537)


lachlandow1 (approved):
More of a question about what the best practices are for storing the analytics id - understand the value is public facing but is it best practice to store on the repo or have as a pipeline variable to keep the configuration separate. 

Approved as doesn't present a risk

PR #3619, opened 2026-04-16
feat(web): added new mapper and removed fields for expedite appeals

musabadam-pins:
## Describe your changes

- Removed Planning Obligation
- Removed design and access statement
- removed plans,drawings and list of plans
- Removed new plans or drawings
- Removed other new supporting documents
- Added new s78 expedite mapper

[Ticket](https://pins-ds.atlassian.net/browse/A2-7821)

### DIFF
--- a/packages/appeals/utils/appeal-type-checks.js
+++ b/packages/appeals/utils/appeal-type-checks.js
@@ -2,6 +2,16 @@ import { APPEAL_CASE_TYPE } from '@planning-inspectorate/data-model';
 import { APPEAL_TYPE } from '../constants/common.js';
 
 /**
+ *
+ * @param {Date} date
+ * @param {Date} afterDate
+ * @returns {boolean}
+ */
+export const dateIsAfterDate = (date, afterDate) => {
+	return date.getTime() >= afterDate.getTime();
+};
+
+/**
  * @typedef {typeof APPEAL_CASE_TYPE['D'] | typeof APPEAL_CASE_TYPE['W']} BaseAppealType
  * @typedef {Record<string, BaseAppealType>} BaseCaseType
  */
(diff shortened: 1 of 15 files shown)


packages/appeals/utils/appeal-type-checks.js
```diff
@@ -34,6 +44,26 @@ export const isExpeditedAppealType = (appealType) => {
 	return Boolean(baseCaseType[appealType] === APPEAL_CASE_TYPE.D);
 };
 
+/**
+ * @param {string | null} appealType
+ * @param {string} caseSubmissionDate
+ * @param {string} applicationDecision
+ * @returns {boolean}
+ */
+export const isS78ExpeditedAppealType = (appealType, caseSubmissionDate, applicationDecision) => {
```
bryn-pins, line 53: should this have a procedure type check and/or feature flag check?
i.e. is it safe to apply expedited mappings to all appeals that meet the below criteria even if the flag is off or the procedure type is not part 1 for any reason?

musabadam-pins: this check only occurs if the appeal type is s78 since expedite doesnt have its own type

musabadam-pins: ive added a check for FF in the places that call the isS78ExpeditedAppealType()

bryn-pins: But it seems possible in the UI currently to switch between Written and Part 1

Image

Image

If someone does that, wouldn't we want the mappings for S78 Written to come through instead of the expedited/part 1 mappings

In my head this check could potentially be (Appeal == S78 && Procedure == Part 1)

Or have I misunderstood something?

musabadam-pins: Spoke to Lachlan about this, at the moment, if procedure type is changed, the form remains the same

musabadam-pins: so currently the only conditions that determine whether something is expedite or not in BO is the appealType, Date and decision

bryn-pins (approved):

PR #3629, opened 2026-04-20
fix(api): link on IP correction notify (A2-7760)

julietteorpen:
## Describe your changes

Small fix to correct links in notifies sent to interested parties when a decision or a correction on a decision is issued.
Both now link directly to IP dashboard.

##Tickets A2-7760 and A2-7797

[Ticket](https://pins-ds.atlassian.net/browse/A2-7760)
and 
[Ticket](https://pins-ds.atlassian.net/browse/A2-7977)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
 	},
 	"dependencies": {
 		"@badeball/cypress-cucumber-preprocessor": "23.2.0",
-		"@planning-inspectorate/data-model": "^2.31.2",
+		"@planning-inspectorate/data-model": "^2.32.0",
 		"cypress": "15.3.0",
 		"got": "^12.5.2",
 		"joi": "^17.7.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
 	},
 	"dependencies": {
 		"@badeball/cypress-cucumber-preprocessor": "23.2.0",
-		"@planning-inspectorate/data-model": "^2.31.2",
+		"@planning-inspectorate/data-model": "^2.32.0",
```
bryn-pins, line 37: should this be part of this PR?

julietteorpen: Tests were failing without updating it, but I can remove if it feels like it shouldn't be here

bryn-pins: Not sure why it would cause issues as 2.32.0 is an applications change but not an issue and given slow tests will approve

bryn-pins (approved):

PR #3664, opened 2026-04-27
feat(infrastructure): update DEV private endpoint config (no-ticket)

Snasim21:
**Describe your changes**

**Issue ticket number and link**

**Type of change 🧩**

 

- [ ]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  Documentation update
- [ ]  Other (please explain in the description section above)
- [ ] 

**Checklist before requesting a review**

- [ ]  I have performed a self-review of my own code
- [ ]  I have double checked this work does not include any hardcoded secrets or passwords
- [ ]  I have made corresponding changes to the documentation
- [ ]  I have provided details on how I have tested my code
- [ ]  I have referenced the ticket number above
- [ ]  I have provided a description of the ticket
- [ ]  I have included unit tests to cover any testable code changes


infrastructure/service-bus.tf
```diff
@@ -56,7 +56,7 @@ resource "azurerm_private_endpoint" "sb_main" {
   private_service_connection {
     name                           = "pins-psc-${local.service_name}-sb-${var.environment}"
     private_connection_resource_id = local.service_bus.id
-    is_manual_connection           = false
+    is_manual_connection           = true
```
beejjacobs: Why do we have this change?

infrastructure/environments/dev.tfvars
```diff
@@ -18,7 +18,7 @@ apps_config = {
   functions_node_version     = 22
   functions_service_plan_sku = "P0v3"
   node_environment           = "development"
-  private_endpoint_enabled   = false
+  private_endpoint_enabled   = true
```
beejjacobs: This is probably the quickest way to do this - but perhaps we could consider replacing the use of this variable with hard-coded true values if all envs will now be true?

open to thoughts on it.

beejjacobs (changes requested):
Looks like the right direction, just a few comments to consider.

infrastructure/environments/dev.tfvars
Snasim21: Good point, Ben. All envs are true now, so I’m happy to hard-code it if we want

PR #3688, opened 2026-05-05
fix(api): return redactedRepresentation with getRepresentations call …

topickering:
…(A2-8021)

## Describe your changes

Add redactedRepresentation to the select statement in getRepresentations call.  This allows calculation of isRedacted status, and makes it redacted rep available when copying representations on unlinking appeals

[Ticket](https://pins-ds.atlassian.net/browse/A2-8021)


appeals/web/environment/config.js
```diff
@@ -125,7 +125,7 @@ export function loadConfig() {
 		cacheControl: {
 			maxAge: CACHE_CONTROL_MAX_AGE || '1d'
 		},
-		googleAnalyticsId: GOOGLE_ANALYTICS_ID || '',
+		googleAnalyticsId: GOOGLE_ANALYTICS_ID || 'id',
```
bryn-pins, line 128: this meant to be here?

topickering: Ah yep, thought I'd add this too as it will prevent the app crashing if running locally without the env var set - defaulting to an empty string doesn't work.

bryn-pins (approved):

PR #3692, opened 2026-05-06
fix(api): broadcast site visit event for child appeals (A2-7986)

topickering:
## Describe your changes

Broadcast site visit events for child appeals.

[Ticket](https://pins-ds.atlassian.net/browse/A2-7986)


zoebutterworth (commented):
Github is having issues adding the comment onto the file. This is for site-visits-delete.test.js approx L90-100

Doesn't the second call of mockResolvedValue overwrite the first? If you want to have it change after one call use mockResolvedValueOnce

zoebutterworth (approved):

PR #3700, opened 2026-05-08
Fix/a2 8017 change lead move reps

topickering:
## Describe your changes

- passes correct params to 'moveRepresentations' function
- broadcasts moved/copied representations on changing lead appeal or unlinking 

[Ticket](https://pins-ds.atlassian.net/browse/A2-8017)

### DIFF
--- a/appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
+++ b/appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
@@ -494,6 +494,7 @@ describe('appeal linked appeals routes', () => {
 						databaseConnector.appealRelationship.create.mockResolvedValue({});
 						// @ts-ignore
 						databaseConnector.appeal.findUnique.mockResolvedValue(testAppeal);
+						databaseConnector.representation.findMany.mockResolvedValue([{ id: 'testId' }]);
 
 						const response = await request
 							.post(`/appeals/${testAppeal.id}/update-linked-appeals`)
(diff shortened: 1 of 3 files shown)


appeals/api/src/server/endpoints/link-appeals/link-appeals.controller.js
```diff
@@ -447,20 +448,30 @@ export const updateLinkedAppeals = async (req, res) => {
...
+		let representationsToBroadcast;
+		// @ts-ignore
+		let representationBroadcastAction;
 
 		switch (operation) {
 			case LINK_APPEALS_CHANGE_LEAD_OPERATION: {
 				await Promise.allSettled([
 					// @ts-ignore
-					moveRepresentations(currentLead.id, appealToReplaceLead.id),
+					(representationsToBroadcast = await moveRepresentations(
```
bryn-pins: Is this correct, await within a Promise.all presumably doesn't run in parallel
Should it be instead

```js
const results = await Promise.allSettled([
					// @ts-ignore
					moveRepresentations(currentLead.id, appealToReplaceLead.id),
					// @ts-ignore
					duplicateAllFiles(currentLead, appealToReplaceLead, options),
					// @ts-ignore
					replaceLeadAppeal(currentLead, appealToReplaceLead)
				]);

representationsToBroadcast = results[0].value;
```

topickering: Have amended to use this approach

appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
zoebutterworth, line 1: Are there not more test cases which should test that the broadcast is happening as expected? It feels like it is also worth checking that the correct variables have been used- especially the update/create as that is dependent on the case

zoebutterworth (approved):

PR #3715, opened 2026-05-14
feat(web): updated written representation display names (a2-8052)

pinsMT:
1. Updated display name applied to:
- Start case > select procedure flow
- Case history procedure type logs
- Case overview procedure type display
- All cases dropdown filter box
2. Util function added for quick access method of changing written rep display names
3. Updates snapshots

Ticket: https://pins-ds.atlassian.net/browse/A2-8052

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
+++ b/appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
@@ -3,6 +3,7 @@ import { getAppealCaseNotes } from '#appeals/appeal-details/case-notes/case-note
 import config from '#environment/config.js';
 import { appealShortReference } from '#lib/appeals-formatter.js';
 import { dateISOStringToDisplayDate, dateISOStringToDisplayTime12hr } from '#lib/dates.js';
+import { getWrittenProcedureTypeDisplayLabel } from '#lib/procedure-type-display-name-formatter.js';
 import { APPEAL_TYPE } from '@pins/appeals/constants/common.js';
 import { utcToZonedTime } from 'date-fns-tz';
 import * as interestedPartyCommentsService from '../representations/interested-party-comments/interested-party-comments.service.js';
@@ -88,6 +89,16 @@ export const renderAudit = async (request, response) => {
 					}
 				});
 			}
+
+			let writtenDisplayName;
+			if (detailsHtml.includes('Part 1')) {
+				writtenDisplayName = getWrittenProcedureTypeDisplayLabel('Part 1');
+				detailsHtml = detailsHtml.replace('Part 1', writtenDisplayName ?? '');
+			} else if (detailsHtml.includes('written')) {
+				writtenDisplayName = getWrittenProcedureTypeDisplayLabel('Written');
+				detailsHtml = detailsHtml.replace('written', writtenDisplayName ?? '');
+			}
+
 			const loggedDate = utcToZonedTime(audit.loggedDate, 'Europe/London');
 			return {
 				dateTime: loggedDate.getTime(),
--- a/appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
+++ b/appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
@@ -3,8 +3,8 @@ import { getAppealCaseNotes } from '#appeals/appeal-details/case-notes/case-note
 import config from '#environment/config.js';
 import { appealShortReference } from '#lib/appeals-formatter.js';
 import { dateISOStringToDisplayDate, dateISOStringToDisplayTime12hr } from '#lib/dates.js';
-import { getWrittenProcedureTypeDisplayLabel } from '#lib/procedure-type-display-name-formatter.js';
-import { APPEAL_TYPE } from '@pins/appeals/constants/common.js';
+import { appealProcedureNameToLabelText } from '#lib/procedure-type-display-name-formatter.js';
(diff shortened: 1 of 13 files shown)


appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
```diff
@@ -88,6 +89,16 @@ export const renderAudit = async (request, response) => {
 					}
 				});
 			}
+
+			let writtenDisplayName;
```
julietteorpen: Only minor but I'm wondering if this variable is really needed as it's only used within the scope of the if statements and not reused between the two conditions?

appeals/web/src/server/lib/procedure-type-display-name-formatter.js
```diff
@@ -0,0 +1,11 @@
...
+ * @param {"Written" | "Part 1" | string} procedureTypeDisplayName
+ * @returns (substring: string, ...args: any[]) => string
+ */
+export const getWrittenProcedureTypeDisplayLabel = (procedureTypeDisplayName) => {
+	if (procedureTypeDisplayName === 'Written') {
+		return 'Written representations (Part 2)';
+	} else if (procedureTypeDisplayName === 'Part 1') {
+		return 'Written representations (Part 1)';
+	}
+};
```
zoebutterworth: I feel like it makes more sense to make this a generic getProcedureTypeDisplayLabel so that you don't need to check the procedure type before calling this

zoebutterworth (changes requested):
Throughout this PR we should be using the key from the procedure type table rather than the name- otherwise changing the names in the database would break all of this logic.

Also if you are using strings for comparison like this then they should be variables somewhere rather than being used all over the place as it makes it harder to change them in the future :)

It looks like there are constants for them in the data model: APPEAL_CASE_PROCEDURE:
https://github.com/Planning-Inspectorate/data-model/blob/af24943374752214da9f08b2081d04e140cfdd99/src/enums.js#L94-L100

appeals/web/src/server/appeals/appeal-details/audit/audit.controller.js
pinsMT: removed in the new commit

appeals/web/src/server/lib/procedure-type-display-name-formatter.js
pinsMT: Removed the first instance of checking procedure type, I've created a place for two functions which receive either the key or the name and map the display name accordingly. I've added a new constant in common.js for comparing display names.

appeals/web/src/server/lib/mappers/data/appeal/submappers/__tests__/case-procedure.mapper.test.js
```diff
@@ -21,7 +20,7 @@ describe('case-procedure.mapper', () => {
 		params = {
 			appealDetails: {
 				startedAt: new Date('2025-01-01'),
-				procedureType: APPEAL_CASE_PROCEDURE.HEARING,
+				procedureType: PROCEDURE_TYPE_DISPLAY_NAME.HEARING,
```
zoebutterworth: For this one I think you do want to keep it as the one from the data model

appeals/web/src/server/lib/procedure-type-display-name-formatter.js
```diff
@@ -0,0 +1,39 @@
+import { capitalizeFirstLetter } from '#lib/string-utilities.js';
+import { PROCEDURE_TYPE_DISPLAY_NAME } from '@pins/appeals/constants/common.js';
+import { APPEAL_CASE_PROCEDURE } from '@planning-inspectorate/data-model';
+
+/**
+ * @param {"Written" | "Part 1" | "Hearing" | "Inquiry" | string} procedureTypeName
```
zoebutterworth: Having the values here will probably not get updated and cause confusion if we are now using the 'PROCEDURE_TYPE_DISPLAY_NAME' values

packages/appeals/constants/common.js
```diff
@@ -87,6 +87,13 @@ export const PROCEDURE_TYPE_MAP = Object.freeze({
 	inquiry: 'an inquiry'
 });
 
+export const PROCEDURE_TYPE_DISPLAY_NAME = Object.freeze({
+	WRITTEN_PART_2: 'Written',
+	WRITTEN_PART_1: 'Part 1',
+	HEARING: 'Hearing',
+	INQUIRY: 'Inquiry'
+});
+
```
zoebutterworth: I'm not sure what it makes sense to call this but I think that calling it the 'procedure type display name' is a little unclear as we aren't actually using this as the value to display- possibly just PROCEDURE_TYPE_NAME as that is what it is in the database?

appeals/web/src/server/lib/mappers/data/appeal/submappers/__tests__/case-procedure.mapper.test.js
pinsMT: agreed to use this value as it matches the typical format the data is received in case-proecdure.mapper

packages/appeals/constants/common.js
pinsMT: Agreed. Name has been changed.

appeals/web/src/server/lib/procedure-type-display-name-formatter.js
pinsMT: removed the unnecessary type def values.

zoebutterworth (approved):

PR #3716, opened 2026-05-14
fix(api): update service user id when moving/copying representations …

topickering:
…(A2-8193, A2-7991)

## Describe your changes

When copying or moving representations on unlinking or changing the lead linked appeal, these changes ensure that the relevant serviceUSerId is set on the copied / moved representations

[Ticket](https://pins-ds.atlassian.net/browse/A2-7991)

### DIFF
--- a/appeals/api/src/server/endpoints/link-appeals/link-appeals.service.js
+++ b/appeals/api/src/server/endpoints/link-appeals/link-appeals.service.js
@@ -7,26 +7,32 @@ import {
 	addDocumentsToAppeal,
 	getFoldersForAppeal
 } from '#endpoints/documents/documents.service.js';
-import addressRepository from '#repositories/address.repository.js';
+import { broadcasters } from '#endpoints/integrations/integrations.broadcasters.js';
 import appealRepository from '#repositories/appeal.repository.js';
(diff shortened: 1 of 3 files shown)


appeals/api/src/server/endpoints/link-appeals/link-appeals.service.js
```diff
@@ -148,23 +152,45 @@ export const unlinkChildAppeal = async (appeal) => {
...
+			}
+
+			representationRepository.updateRepresentationById(representation.id, updateData);
+		})
 	);
+
+	const { comments: representationsInDestinationAppeal } =
+		await representationRepository.getRepresentations([Number(destinationAppeal.id)]);
+
+	return representationsInDestinationAppeal;
```
zoebutterworth: I don't think this is consistent with what it was doing previously. Before it used updateRepresentations which returned the reps from the repIds which had been updated. If there is a scenario where there are reps on the child (not the case for multiple appellant) already then this will return the reps of those as well as the ids of the changed reps

topickering: Good point! Have amended so only returns the successfully updated representations

musabadam-pins (approved):

PR #3735, opened 2026-05-22
fix(web): bulk download (a2-7463, a2-7914)

bryn-pins:
## Describe your changes

- extends request timeout for bulk download
- blob download streams to work in a pipeline, download starts when the previous file has finished downloading and started archiving to avoid the blob stream itself timing out
- reduces compression level - excessive cpu usage for little gain (skipped entirely for pre compressed files)
- includes internal folders in the download

https://pins-ds.atlassian.net/browse/A2-7463
https://pins-ds.atlassian.net/browse/A2-7914

### DIFF
--- a/appeals/web/src/server/app/components/file-downloader.component.js
+++ b/appeals/web/src/server/app/components/file-downloader.component.js
@@ -5,6 +5,8 @@ import {
 	getFileVersionsInfo,
 	getRepresentationAttachments
 } from '#appeals/appeal-documents/appeal.documents.service.js';
+import getActiveDirectoryAccessToken from '#lib/active-directory-token.js';
+import logger from '#lib/logger.js';
 import { camelCaseToWords, toSentenceCase } from '#lib/string-utilities.js';
 import { BlobServiceClient } from '@azure/storage-blob';
 import config from '@pins/appeals.web/environment/config.js';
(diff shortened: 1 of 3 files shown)


appeals/web/src/server/app/components/file-downloader.component.js
```diff
@@ -401,75 +491,59 @@ export const getBulkFileInfo = async (apiClient, caseId, representationType) =>
...
-									representationAttachmentFullName || `${folderPath}/${document.name}`,
-								blobStorageContainer,
-								blobStoragePath,
-								documentURI
-							};
+	const bulkFileInfo = folders.flatMap((folder) => {
+		let folderPath = folder.path
+			.split('/')
+			.map((folderName) => camelCaseToWords(folderName.trim()).replace('Lpa', 'LPA'))
+			.map((folderName) => (folderName === 'Internal' ? 'Case Management (Internal)' : folderName))
```
zoebutterworth: Do we have a constant for 'Internal' rather than putting it as a magic string here?

appeals/web/src/server/app/components/file-downloader.component.js
```diff
@@ -236,6 +255,70 @@ const createBlobDownloadStream = async (
...
+const addBlobsToArchive = async (archive, blobStorageClient, bulkFileInfo, missingFiles) => {
+	let nextDownloadPromise = null;
+	for (let i = 0; i < bulkFileInfo.length; i++) {
+		const fileInfo = bulkFileInfo[i];
+
+		let downloadPromise;
+		if (nextDownloadPromise) {
+			downloadPromise = nextDownloadPromise;
+			nextDownloadPromise = null;
+		} else {
```
zoebutterworth, line 273: Could this part cause issues if nextDownloadPromise has returned null before we reach this? I guess it would just run createBlobDownloadStream twice for the same file

bryn-pins: I don't think so it's the promise object itself not the result from the promise until awaited, so think will always be an object

zoebutterworth (approved):

PR #3739, opened 2026-05-26
[A2 1234] update and fix hearing tests

marcpodgorney81:
## Describe your changes 

This PR contains various fixes and updates 
- hearing related tests (which required updating following a change that removed the optional 'Do you know date and time question') 
- document related tests (i.e. manage docs) 
- other functionality and tests that discovered wasn't working!   

**For hearing tests** 

- fixed the startS78HearingCase function in happyPathHelper, which starts a case as hearing 
- refactoring of the hearing.spec file 
   - updating it so that creates test case per scenario (was using a single test case across all which think was causing some issues, or at least could lead to some) 
   - adding and refactoring some scenarios 
   - updating test setup to use the advanceTo function, to make more efficient 
- adding some helper functions 
      - hearingSectionPage 
      - caseHistoryPage 
      - happyPathHelper    

**For manage docs/correspondance tests** 
These were failing intermittently, due to virus scan blocking ability to view a document e.g. To make them more stable and consistent have added the `cy.simulateDocumentsScanComplete` command, this has the effect of marking as scanned any document associated with the specified case  

### Some other fixes and updates

**added missing annotations** 
Some files were missing this annotations  
```
// @ts-nocheck 
/// <reference types="cypress"/>
``` 

**added an object to track cta text**
One of the causes of tests failing was changes to cta text. In order to make this easier to manage I've created `consts.js` with an object to keep track of some common cta labels, such that if there is a change we only need to update in the one file 

**added logs to login command**
Given that are planning to add some tests for inspector user, thought it would be good to add some logs to give visibility of which user type are logging in as. 

**new command to check notifies not sent** 
When adding a new scenario for not cancelling hearing, realised would be good to have an 'opposite' of `checkNotifesSent` which checks that specified notified were not sent. Another use case for this (as possible follow up) could be to check that notifies are not sent when setting up site visit without date e.g. 

**fixed logging issue in appealsApiClient file** 
I noticed that in event of an error when making an api call, the message wasn't being logged. This was due to issue in formatting 

**updated case officer Id (i.e. for assigning case officer via api)**
Found that although this api call worked, it did not assign an officer to the case. This seems to be because the id has been updated (did a manual assign and checked id using swagger)? Have updated Id and moved to users.js file as seemed logical place for it?   

**updated advanceTo function**
- added some comments to help explain the logic 
- extra parameter to determine whether to load case details as last step 
- check for if hearing exists as part of EVENT_READY_TO_SETUP actions 

*{No ticket - fixes}*


appeals/e2e/cypress/e2e/back-office-appeals/manageDocsAppellant.spec.js
```diff
@@ -153,43 +170,39 @@ describe('manage docs on appellant case', () => {
...
-					caseDetailsPage.clickButtonByText('Add appellant proof of evidence and witness');
-
-					// check page header and success banner
-					caseDetailsPage.validateSectionHeader('Review appellant proof of evidence and witnesses');
-					caseDetailsPage.validateBannerMessage(
-						'Success',
-						'Appellant proof of evidence and witnesses added'
-					);
-				});
+	it('can upload appellant proof of evidence and witness to inquiry case', () => {
```
marcpodgorney81, line 173: fyi this was just me removing the smoke tag

jacksolirius (commented):
LGTM

jacksolirius (approved):

PR #3744, opened 2026-05-27
chore(appeals): moving string case utils out to top level utils from …

zoebutterworth:
…web and api (no-ticket)

## Describe your changes

Moving string case utils out of web and api and adding them to the top level utils

Have checked by grepping recursively for '#lib/string-utilities' and '#utils/string-utils' to check that all imports are updated. This is also checked by tscheck

No ticket


zoebutterworth:
opted out of e2e tests- only the known issue with hearing test failed

bryn-pins (approved):

PR #3749, opened 2026-05-28
fix(appeals): updating elb with planning-permission-granted to displa…

zoebutterworth:
…y correct text (A2-8107)

+ appellant pulled from child appeal on child appeal page

## Describe your changes

- Creating a common decisionOutcomeToDisplayText function so that ELB planning-permission-granted is displayed as 'Listed building consent granted'
- Updating the various locations which display the decision outcome to use that function
- Adding tests for these specific cases
- Updating to pull the appellant from the child appeal on the child appeal page

## Local testing

Tested that for a single appeal, lead appeal and child appeal the decision outcome displays correctly for both an enforcement appeal and an ELB appeal
Tested that the issue decision flow works correctly for a lead and child appeal which are not both ELB (both ways around- child ELB or lead ELB)

[Ticket](https://pins-ds.atlassian.net/browse/A2-8107)

### DIFF
--- a/appeals/api/src/server/endpoints/decision/__tests__/decision.test.js
+++ b/appeals/api/src/server/endpoints/decision/__tests__/decision.test.js
@@ -256,6 +256,14 @@ describe('decision routes', () => {
 				'invalid'
 			],
 			[
+				'enforcementNoticeAppeal',
+				'planning_permission_granted',
+				'Because it is.',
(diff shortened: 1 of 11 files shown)


zoebutterworth:
Adding e2e opt out label as the hearing test is failing (ongoing known issue)

topickering (approved):

PR #3774, opened 2026-06-04
fix(api): update documents when changing lead (A2-8321)

topickering:
## Describe your changes

Representations get moved to the new lead when the lead appeal is changed for linked appeals.  However, documents attached to those reps retained the previous lead appealId and folderId.  This PR amends this so the documents relating to any moved reps are also updated, and the updated rep is broadcast to FO.

[Ticket](https://pins-ds.atlassian.net/browse/A2-8321)

### DIFF
--- a/appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
+++ b/appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
@@ -489,13 +489,20 @@ describe('appeal linked appeals routes', () => {
 							}
 						];
 						testAppeal.appellantCase.appellantCaseValidationOutcome = { id: 1, name: 'Valid' };
-						databaseConnector.folder.findMany.mockResolvedValue([]);
+						databaseConnector.folder.findMany.mockResolvedValue([{ id: 1 }]);
 						databaseConnector.appealRelationship.deleteMany.mockResolvedValue({});
 						databaseConnector.appealRelationship.create.mockResolvedValue({});
 						// @ts-ignore
 						databaseConnector.appeal.findUnique.mockResolvedValue(testAppeal);
 						databaseConnector.representation.findMany.mockResolvedValue([{ id: 'testId' }]);
 						databaseConnector.representation.update.mockResolvedValue({ id: 'testId' });
+						databaseConnector.representationAttachment.findMany.mockResolvedValue([
+							{ documentGuid: 'testDocGuid' }
+						]);
+						databaseConnector.document.updateMany.mockResolvedValue([]);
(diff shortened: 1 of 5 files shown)


appeals/api/src/server/repositories/representation.repository.js
```diff
@@ -307,6 +307,42 @@ const addAttachments = async (repId, attachments) => {
...
+ * @param {number} destinationFolderId
+ */
+const moveRepresentationAttachmentDocuments = async (
+	repId,
+	destinationAppealId,
+	destinationFolderId
+) => {
+	const attachmentsToUpdate = await databaseConnector.representationAttachment.findMany({
+		where: { representationId: repId }
+	});
```
zoebutterworth, line 322: Do you need to do this one if you are going to do the findMany on the return? Could you use the representationId on both of those calls?

topickering: I think we do need it but could be wrong - documents don't store a repId and reps down't store a docId/guid, so we need to go via the repAttachment table for the first call in order to get the guids needed to identify which docs to update.  It would be possible just to do a call on the rep table and include the Attachments, but I'm not sure if that is any more efficient (given it would be prisma doing some stuff under the hood).

The final return call gets the docs from the document table and I'm not sure how to get them directly from the repId. Could do a very nested call (repeatable_>include attachements>docuemntVersion>document(from latest version) but I 'm not overly keen on making a call nested to those levels.

appeals/api/src/server/endpoints/link-appeals/__tests__/link-appeal.test.js
```diff
@@ -489,13 +489,21 @@ describe('appeal linked appeals routes', () => {
 							}
 						];
 						testAppeal.appellantCase.appellantCaseValidationOutcome = { id: 1, name: 'Valid' };
-						databaseConnector.folder.findMany.mockResolvedValue([]);
+						databaseConnector.folder.findMany.mockResolvedValue([{ id: 1 }]);
+						// databaseConnector.folder.findMany.mockResolvedValue([]);
```
zoebutterworth: Did you mean to leave this comment in?

topickering: Ah, no I didn't! Will remove

appeals/api/src/server/repositories/representation.repository.js
zoebutterworth: Following further discussion I am happy for this to remain as it is :)

zoebutterworth (approved):

PR #3782, opened 2026-06-04
fix(api): amend default caseStartedDate hour to 10am from 00 (A2-7916)

topickering:
## Describe your changes

Uses the DAYTIME_HOUR constant when setting caseStarted date - this is time corrected and prevents cases started in BST having a caseStartedDate of 23:00 the day before.

caseStartedDate is also added to the relevant audit trail message.

NOTE - this does **not** amend the calculation of the appeal timetable.

[Ticket](https://pins-ds.atlassian.net/browse/A2-7916)


topickering:
Closing PR without merging as agreed to retain 00:00 for case start date

PR #3784, opened 2026-06-05
feat(appeals): ingest and display reasonForAppealAppellant (A2-8391)

lachlandow1:
## Describe your changes

Adds logic to ingest and map the appellant reason for appeal question from FO.

Add tests to cover displaying new question

[Ticket](https://pins-ds.atlassian.net/browse/A2-8391)


appeals/web/src/server/appeals/appeal-details/appellant-case/__tests__/appellant-case-expedited.test.js
zoebutterworth, line 1: These could be combined into one block using an it.each to avoid some repetition

zoebutterworth (approved):
Not a blocking comment :)

PR #3801, opened 2026-06-09
feat(web): changing bys rows for enforcement and elb appellant case (…

zoebutterworth:
…A2-8062)

## Describe your changes

- Adding logic to remove '' and '' from enforcement and elb based on the status of the new BYS feature flag
- Adding 'What is your appeal about?' to enforcement and elb based on the status of the new BYS feature flag
- Adding logic to the displaying of 'What is your appeal about?' to display the correct text for enforcement/elb
- Updating tests to not have those parameters true for the tests which are not setting them specifically
- Updating snaps to cover this change

Locally tested running the following e2e test suites (none failing):
- appellantCase
- endToEndEnforcement
- endToEndLinkedEnforcement

Locally checked that the correct text was showing for all appeal types:
- Enforcement and ELB show updated version
image
image

- Other appeal types show no change

Locally checked that changing the feature flag to false returns to the previous behaviour for all appeal types

[Ticket](https://pins-ds.atlassian.net/browse/A2-8062)


topickering (approved):
LGTM

PR #3808, opened 2026-06-10
fix(web): use default deadline values when change procedure timetable…

topickering:
… (A2-8223)

When the appeal procedure is changed, the recalculated timetable defaults to 00:00 for hours/minutes.  This can cause the final day to be cut short.  This change ensures that the recalculated timetable on change procedure uses the default deadline hours/minutes (ie 23:59, timezone adjusted to UTC)

[Ticket](https://pins-ds.atlassian.net/browse/A2-8223)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/change-procedure-type/change-procedure-check-and-confirm/__tests__/change-procedure-check-and-confirm.test.js
+++ b/appeals/web/src/server/appeals/appeal-details/change-procedure-type/change-procedure-check-and-confirm/__tests__/change-procedure-check-and-confirm.test.js
@@ -720,12 +720,12 @@ describe('GET /change-appeal-procedure-type/check-and-confirm', () => {
 				existingAppealProcedure: 'hearing',
 				appealProcedure: 'inquiry',
 				eventDate: '3025-02-01T12:00:00.000Z',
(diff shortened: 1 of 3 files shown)


zoebutterworth:
I had a look in the codebase at how these are normally set (when you start the case the timetable is set with no issues) and it uses `setTimeInTimeZone(calculatedDate, DEADLINE_HOUR, DEADLINE_MINUTE)` from here: https://github.com/Planning-Inspectorate/appeals-back-office/blob/main/packages/appeals/utils/business-days.js#L214C1-L234C2. It feels like this function should be used if possible for consistency

topickering:
> I had a look in the codebase at how these are normally set (when you start the case the timetable is set with no issues) and it uses `setTimeInTimeZone(calculatedDate, DEADLINE_HOUR, DEADLINE_MINUTE)` from here: https://github.com/Planning-Inspectorate/appeals-back-office/blob/main/packages/appeals/utils/business-days.js#L214C1-L234C2. It feels like this function should be used if possible for consistency

I've updated to use that function.  That one errors if there is not date (unlike the previous function, which returns an empty string) so I've added some ternary checks.

zoebutterworth (approved):

PR #3848, opened 2026-06-23
fix(web): removed mapper from data chain and removed mapping command (a2-8459)

pinsMT:
- Removed mapping command
- Removed passed data for design and access upload mapper
- Updated snap

Ticket: https://pins-ds.atlassian.net/browse/A2-8458


appeals/web/src/server/appeals/appeal-details/appellant-case/page-components/cas.mapper.js
lachlandow1, line 1: The design and access statement is still available on CAS appeals before the expedited 1st april cut off - we need to make the display of the field conditional based on the date

PR #3849, opened 2026-06-23
fix(web): removed mapping commands, updated tests and snaps (a2-8460)

pinsMT:
- Removed mapping commands
- Removed data passed to cas advert appellant case maps
- Updated snaps
- Updated tests

Ticket: https://pins-ds.atlassian.net/browse/A2-8460


appeals/web/src/server/appeals/appeal-details/appellant-case/page-components/cas-advert.mapper.js
lachlandow1, line 1: We should also check for the cutoff date here. As this can also still be displayed if before 1st april submission date

pinsMT: restored the option an added the condtion

lachlandow1 (dismissed):

lachlandow1 (approved):

PR #3860, opened 2026-06-25
fix(web): added condition to hide upload option (a2-8458)

pinsMT:
- Added condition in CAS appellant case mapper which hides upload option for cases after application cut off date  
- Added test

Ticket: https://pins-ds.atlassian.net/browse/A2-8458


appeals/web/src/server/appeals/appeal-details/appellant-case/page-components/cas.mapper.js
lachlandow1, line 1: Would be good to add tests for this

lachlandow1:
Would also be good to update the commit message for this to describe what has changed closer to the ticket description

appeals/web/src/server/appeals/appeal-details/appellant-case/page-components/cas.mapper.js
lachlandow1: Looks good

lachlandow1 (approved):

PR #3866, opened 2026-06-26
fix(appeals): remove statement of common ground from S78 expedited appeal form and PDF (A2-8599)

lachlandow1:
## Describe your changes

Remove statement of common ground from the S78 appeal form and PDF generator

Ensured unit tests passed and added new unit tests for logic change
Tested in browser

[Ticket](https://pins-ds.atlassian.net/browse/A2-8599)

### DIFF
--- a/appeals/pdf-service-api/src/mappers/appellant-case/__tests__/__snapshots__/appellant-case.mapper.test.js.snap
+++ b/appeals/pdf-service-api/src/mappers/appellant-case/__tests__/__snapshots__/appellant-case.mapper.test.js.snap
@@ -1228,10 +1228,6 @@ exports[`mapAppellantCaseData appellantCase should map appellant case data for a
         },
         {
           "html": "No documents",
-          "key": "Draft statement of common ground",
-        },
-        {
-          "html": "No documents",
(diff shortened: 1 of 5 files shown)


appeals/web/src/server/appeals/appeal-details/appellant-case/__tests__/__snapshots__/appellant-case-main.test.js.snap
```diff
@@ -3908,7 +3908,7 @@ exports[`appellant-case-main GET /appellant-case should render the appellant cas
                 <div class="govuk-summary-card__content">
                     <dl class="govuk-summary-list" id="appeal-summary">
                         <div class="govuk-summary-list__row"><dt class="govuk-summary-list__key"> How would you prefer us to decide your appeal?</dt>
-                            <dd                             class="govuk-summary-list__value">Not answered</dd>
+                            <dd                             class="govuk-summary-list__value">Hearing</dd>
```
bryn-pins: I think this has changed on main since you ran and snapshots need rerunning, seems to be failing in tests and can't see why these changes would appear here

bryn-pins (approved):

PR #3873, opened 2026-06-30
fix(web): validate API parameters that come from request parameters (no-ticket)

beejjacobs:
## Describe your changes

Ensure API requests from the web app which use request parameters are safe, checking parameters are valid before making the network request.

The majority of this code was written by GitHub Copilot, but I have checked it over and validated the tests.

https://pins-ds.atlassian.net/browse/A2-8645

### DIFF
--- a/appeals/web/src/server/appeals/appeal-documents/appeal.documents.service.js
+++ b/appeals/web/src/server/appeals/appeal-documents/appeal.documents.service.js
@@ -1,5 +1,6 @@
 import logger from '#lib/logger.js';
 import { fileType } from '#lib/nunjucks-filters/mime-type.js';
+import { assertValidIds, assertValidNumericIds } from '#lib/validators/api-parameters.validator.js';
 
 /** @typedef {import('@pins/appeals.api').Appeals.FolderInfo} FolderInfo */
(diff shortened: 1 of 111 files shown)


appeals/web/src/server/appeals/appeal-documents/appeal.documents.service.js
```diff
@@ -140,8 +146,10 @@ export const getDocumentRedactionStatuses = async (apiClient) => {
 
 export const updateDocument = async (apiClient, appealId, documentDetail) => {
 	try {
+		const ids = assertValidNumericIds({ appealId });
+		assertValidIds({ id: documentDetail.document.id });
```
bryn-pins: Bit of a nitpick but it seems a bit odd to me to return the params back from assertValidNumericIds and use that in the uri but not do the same with assertValidIds
Don't have a strong feeling about which is better as it just throws anyway

beejjacobs: I think this was for typing (since the assertion returns a narrower type). I will go with using the returned value.

beejjacobs:
I've rebased and resolved conflicts, and ensured a consistent approach of using the return values from the assertion.

bryn-pins (approved):

PR #3881, opened 2026-07-01
fix(appeals): update date comparison function to cover on the day changes (A2-8662)

lachlandow1:
## Describe your changes

Updated date function to to convert date time to yyyy-MM-dd for comparison.
Covers issue where 1st April was not included in date range for the expedited cutoff check.

Added unit tests to cover new function logic
Manually tested in browser

[Ticket](https://pins-ds.atlassian.net/browse/A2-8662)

### DIFF
--- a/packages/appeals/utils/appeal-type-checks.js
+++ b/packages/appeals/utils/appeal-type-checks.js
@@ -7,15 +7,7 @@ import {
 import { APPEAL_TYPE, PROCEDURE_TYPE_NAME } from '../constants/common.js';
 import { EXPEDITED_ORIGINAL_APPLICATION_CUTOFF } from '../constants/dates.js';
 
-/**
- *
- * @param {Date} date
- * @param {Date} afterDate
- * @returns {boolean}
- */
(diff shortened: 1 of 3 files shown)


packages/appeals/utils/appeal-type-checks.js
```diff
@@ -79,7 +71,7 @@ export const isS78ExpeditedAppealType = (
 	if (!appealType || !caseSubmissionDate) return false;
 
 	const isS78 = appealType === APPEAL_CASE_TYPE.W || appealType === APPEAL_TYPE.S78;
-	const isAfterCutoff = dateIsAfterDate(
+	const isAfterCutoff = dateIsOnOrAfterDate(
```
zoebutterworth: Do you feel we should use !beforeExpeditedOriginalApplicationCutOff rather than repeating the logic here?

lachlandow1: yeah, good spot, I've updated to use beforeExpeditedOriginalApplicationCutOff

zoebutterworth (approved):

PR #3886, opened 2026-07-01
fix(web): stop routing to env services review page from lpaq if already set (a2-8463)

pinsMT:
- Set environmental impact assessment screening value at validation stage
- Added screening value as condition in LPAQ stage - bypassing the routing if value was previously set
- Added tests

Ticket: https://pins-ds.atlassian.net/browse/A2-8463


musabadam-pins (approved):

appeals/web/src/server/appeals/appeal-details/lpa-questionnaire/__tests__/lpa-questionnaire.test.js
```diff
@@ -2139,6 +2139,74 @@ describe('LPA Questionnaire review', () => {
 				);
 			}
 		);
+
+		it.each([
+			['S78', APPEAL_TYPE.S78],
+			['S20', APPEAL_TYPE.PLANNED_LISTED_BUILDING],
```
lachlandow1, line 2145: I think this may have only needed to test S78 expedited as the other appeal types use different appeal forms

PR #3895, opened 2026-07-07
feat(api): update state machine transition for enfLdcDisc hearing (A2…

topickering:
…-7383)

Update state machine:
- new ctx field - isLdcEnfdiscontinuance, defaults to false.
- amend trasitions from statements and final comments to cater for hearings for isLdcEnfDisc

NOTE - this should also be useful for inquiries for ldc/enf/disc.

[Ticket](https://pins-ds.atlassian.net/browse/A2-7383)

### DIFF
--- a/appeals/api/src/server/state/__tests__/state-machine.test.js
+++ b/appeals/api/src/server/state/__tests__/state-machine.test.js
@@ -17,15 +17,15 @@ import createStateMachine from '../create-state-machine';
  * Function to get the next state for HAS appeals based on the initial state and event.
  * @param {string} initial - The initial state of the appeal.
  * @param {string} event - The event that triggers the state transition.
- * @param {boolean} [siteVisitElapsed] - Whether the site visit has elapsed.
+ * @param {boolean} [eventElapsed] - Whether the event has elapsed.
  * @return {import('xstate').StateValue} - The next state of the appeal.
  **/
-const nextStateHAS = (initial, event, siteVisitElapsed = false) => {
+const nextStateHAS = (initial, event, eventElapsed = false) => {
 	const machine = createStateMachine(
 		APPEAL_CASE_TYPE.D,
 		APPEAL_CASE_PROCEDURE.WRITTEN,
 		initial,
-		siteVisitElapsed
+		eventElapsed
 	);
 	const service = interpret(machine).start();
 
@@ -38,11 +38,24 @@ const nextStateHAS = (initial, event, siteVisitElapsed = false) => {
  * @param {string} initial - The initial state of the appeal.
  * @param {string} event - The event that triggers the state transition.
  * @param {string} procedureType - The event that triggers the state transition.
- * @param {boolean} [siteVisitElapsed] - Whether the site visit has elapsed.
+ * @param {boolean} [eventElapsed] - Whether the event has elapsed.
+ * @param {boolean} [ldcEnforcementDiscontinuance] - Whether the case type is ldc, enforcement, elb or discontinuance
  * @return {import('xstate').StateValue} - The next state of the appeal.
  **/
-const nextStateFPA = (initial, event, procedureType, siteVisitElapsed = false) => {
-	const machine = createStateMachine(APPEAL_CASE_TYPE.W, procedureType, initial, siteVisitElapsed);
+const nextStateFPA = (
+	initial,
+	event,
+	procedureType,
+	eventElapsed = false,
+	ldcEnforcementDiscontinuance = false
+) => {
+	const machine = createStateMachine(
+		APPEAL_CASE_TYPE.W,
+		procedureType,
+		initial,
+		eventElapsed,
+		ldcEnforcementDiscontinuance
+	);
 	const service = interpret(machine).start();
 
 	service.send(event);
@@ -324,7 +337,7 @@ describe('State Machine Transitions', () => {
 				APPEAL_CASE_STATUS.INVALID
 			]
 		])(
-			'correctly transitions from %s state on %s event to %s state for FPA',
+			'correctly transitions from %s state on %s event to %s state for FPA - non ldcEnfDiscontinuance',
 			(initial, event, expectedFPAWritten, expectedFPAHearing, expectedFPAInquiry) => {
 				expect(nextStateFPA(initial, event, APPEAL_CASE_PROCEDURE.WRITTEN)).toBe(
 					expectedFPAWritten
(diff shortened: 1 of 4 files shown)


appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -39,10 +39,23 @@ const nextStateHAS = (initial, event, siteVisitElapsed = false) => {
  * @param {string} event - The event that triggers the state transition.
  * @param {string} procedureType - The event that triggers the state transition.
  * @param {boolean} [siteVisitElapsed] - Whether the site visit has elapsed.
+ * @param {boolean} [ldcEnforcementDiscontinuance] - Whether the case type is lpdc, enforcement or discontinuance
```
zoebutterworth: lpdc typo. Also it is probably best to include elb for clarity (or something to make it clear enforcement means enforcement or elb)

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -337,6 +350,58 @@ describe('State Machine Transitions', () => {
...
+			],
+			[
+				APPEAL_CASE_STATUS.STATEMENTS,
+				VALIDATION_OUTCOME_INVALID,
+				APPEAL_CASE_STATUS.INVALID,
+				APPEAL_CASE_STATUS.INVALID,
+				APPEAL_CASE_STATUS.INVALID
+			]
+		])(
+			'correctly transitions from %s state on %s event to %s state for FPA - non ldcEnfDiscontinuance',
```
zoebutterworth: Shouldn't this be ldcEnfDiscontinuance rather than non ldcEnfDiscontinuance?

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -384,7 +472,21 @@ describe('State Machine Transitions', () => {
 			expect(nextStateFPA(initialState, event, procedure, siteVisitElapsed)).toBe(expectedState);
 		});
 
-		test('should transition to ISSUE_DETERMINATION if site visit HAS elapsed', () => {
+		test('should transition to AWAITING_EVENT if site visit has NOT elapsed - hearing ldcEnfDisc', () => {
```
zoebutterworth: 'should transition to AWAITING_EVENT if hearing has NOT elapsed - hearing ldcEnfDisc' rather than site visit for clarity, or change this and the above to refer to the event rather than the specific type?

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -39,10 +39,23 @@ const nextStateHAS = (initial, event, siteVisitElapsed = false) => {
...
+ * @param {boolean} [ldcEnforcementDiscontinuance] - Whether the case type is ldc, enforcement, elb or discontinuance
  * @return {import('xstate').StateValue} - The next state of the appeal.
  **/
-const nextStateFPA = (initial, event, procedureType, siteVisitElapsed = false) => {
-	const machine = createStateMachine(APPEAL_CASE_TYPE.W, procedureType, initial, siteVisitElapsed);
+const nextStateFPA = (
+	initial,
+	event,
+	procedureType,
+	siteVisitElapsed = false,
```
zoebutterworth: This should probably be eventElapsed rather than siteVisitElapsed

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -384,7 +472,21 @@ describe('State Machine Transitions', () => {
 			expect(nextStateFPA(initialState, event, procedure, siteVisitElapsed)).toBe(expectedState);
 		});
 
-		test('should transition to ISSUE_DETERMINATION if site visit HAS elapsed', () => {
+		test('should transition to AWAITING_EVENT if site visit has NOT elapsed - hearing ldcEnfDisc', () => {
+			const initialState = APPEAL_CASE_STATUS.FINAL_COMMENTS;
+			const event = VALIDATION_OUTCOME_COMPLETE;
+			const procedure = APPEAL_CASE_PROCEDURE.HEARING;
+			const siteVisitElapsed = false;
```
zoebutterworth: Clearer if this says eventElapsed rather than siteVisitElapsed

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -384,7 +472,21 @@ describe('State Machine Transitions', () => {
...
+			const isLdcEnfDisc = true;
+
+			const expectedState = APPEAL_CASE_STATUS.EVENT;
```
zoebutterworth, line 486: Based on the title of this test should this be AWAITING_EVENT? i.e. hearing has been set up but hasn't elapsed

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -395,6 +497,20 @@ describe('State Machine Transitions', () => {
 			expect(nextStateFPA(initialState, event, procedure, siteVisitElapsed)).toBe(expectedState);
 		});
 
+		test('should transition to ISSUE_DETERMINATION if site visit HAS elapsed - hearing ldcEnfDisc', () => {
```
zoebutterworth: Same as the comments on the equivalent- we should probably refer to the 'site visit' as event or hearing

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -432,7 +548,7 @@ describe('State Machine Transitions', () => {
 				APPEAL_CASE_STATUS.INVALID
 			]
 		])(
-			'correctly remains at %s state for HAS and FPA - hearing and inquiry',
+			'correctly remains at %s state for HAS and FPA - hearing (non-ldcEnfDisc) and inquiry',
```
zoebutterworth: Do we need an equivalent test suite for ldcEnfDisc?

appeals/api/src/server/state/create-state-machine.js
```diff
@@ -436,16 +442,53 @@ const targetStateOnEventCancelled = {
...
+	switch (procedureType) {
+		case APPEAL_CASE_PROCEDURE.HEARING: {
+			return isLdcOrDiscontinuanceOrEnforcementCaseType
+				? APPEAL_CASE_STATUS.FINAL_COMMENTS
+				: APPEAL_CASE_STATUS.EVENT;
+		}
+		case APPEAL_CASE_PROCEDURE.INQUIRY:
+			return APPEAL_CASE_STATUS.EVIDENCE;
+		case APPEAL_CASE_PROCEDURE.WRITTEN:
+			return APPEAL_CASE_STATUS.FINAL_COMMENTS;
```
zoebutterworth: It doesn't matter too much but I would put these in the order, written, hearing then inquiry as I think that is consistent with the order we normally quote them

appeals/api/src/server/state/create-state-machine.js
```diff
@@ -15,26 +15,33 @@ import {
...
- * @typedef {import('@planning-inspectorate/data-model').APPEAL_CASE_TYPE} AppealType
+ * @typedef {import('@planning-inspectorate/data-model').APPEAL_CASE_TYPE} AppealTypeKey
  * @typedef {import('@planning-inspectorate/data-model').APPEAL_CASE_PROCEDURE} ProcedureType
  * @typedef {import('@planning-inspectorate/data-model').APPEAL_CASE_STATUS} CaseStatus
  * @typedef {object} AdditionalCheckObject
  * */
 
 /**
- * @param {AppealType} appealType
+ * @param {AppealTypeKey} appealTypeKey NOTE - at present implementations only pass 'W' or 'D' to this function
```
zoebutterworth, line 25: As you changed to calling this normalizedAppealTypeKey in transition-state is there a specific reason why you still want it to be appealTypeKey here?

topickering: I wanted the stateMachine function to be able to take any appealTypeKey - the fact we normalise it in implementation shouldn't change that.  If we do want to limit the keys which can be passed to the function then I'd want to update the param type

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: So for these tests the hearings/inquries should not be in FC state so won't move on, whereas the moving on is covered in the tests above (i.e. for ldcEnfDisc it shouldn't remain in that state, so these tests aren't required)

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: Good spot

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: fixed - title was out of date

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: fixed

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: good spot

topickering: amended

appeals/api/src/server/state/__tests__/state-machine.test.js
topickering: will fix

appeals/api/src/server/state/create-state-machine.js
topickering: Fixed

appeals/api/src/server/state/__tests__/state-machine.test.js
```diff
@@ -395,6 +504,24 @@ describe('State Machine Transitions', () => {
 			expect(nextStateFPA(initialState, event, procedure, siteVisitElapsed)).toBe(expectedState);
 		});
 
+		test('should transition to ISSUE_DETERMINATION if hearing HAS elapsed - hearing ldcEnfDisc', () => {
```
zoebutterworth: You have added inquiry to the test but not to the name

topickering: Oops! Updated

zoebutterworth (approved):

PR #3896, opened 2026-07-09
feat(web): adding final comments to ldc, enf, elb, disc hearing and i…

zoebutterworth:
…nquiry (A2-8533)

## Describe your changes

- Reordering the rows so that the hearing/inquiry row is at the bottom of the timetable section
- Removing statement of common ground for enforcement, ELB, discontinuance and LDC hearing
- Change timetable dates shows only the correct dates based on appeal type and procedure based on the above changes
- Updating tests to cover the above changes
- Created new function displayFinalComments in business-rules and replaced logic in various places with this
- Test for displayFinalComments
- Updating appeal-type-checks to use the mappers
- Adding some tests for appeal-type-checks

Tested locally:
image
image

with planning obligation
image
image

[Ticket](https://pins-ds.atlassian.net/browse/A2-8533)

### DIFF
--- a/appeals/web/src/server/lib/mappers/data/appeal/submappers/lpa-final-comments.mapper.js
+++ b/appeals/web/src/server/lib/mappers/data/appeal/submappers/lpa-final-comments.mapper.js
@@ -9,7 +9,7 @@ import {
 	mapRepresentationDocumentSummaryActionLink
 } from '#lib/representation-utilities.js';
 import { APPEAL_REPRESENTATION_STATUS } from '@pins/appeals/constants/common.js';
-import { doNotDisplayFinalComments } from '../common.js';
+import { displayFinalComments } from '@pins/appeals/utils/business-rules.js';
 
 /** @type {import('../mapper.js').SubMapper} */
 export const mapLPAFinalComments = ({ appealDetails, currentRoute, request }) => {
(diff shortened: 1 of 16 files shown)


appeals/web/src/server/lib/mappers/data/appeal/submappers/lpa-final-comments.mapper.js
```diff
@@ -9,7 +9,7 @@ import {
 	mapRepresentationDocumentSummaryActionLink
 } from '#lib/representation-utilities.js';
 import { APPEAL_REPRESENTATION_STATUS } from '@pins/appeals/constants/common.js';
-import { doNotDisplayFinalComments } from '../common.js';
+import { displayFinalComments } from '@pins/appeals/utils/business-rules.js';
```
topickering, line 12: I like this, better to have the positive case than a do not display function

packages/appeals/utils/business-rules.js
```diff
@@ -0,0 +1,12 @@
+import { APPEAL_CASE_PROCEDURE } from '@planning-inspectorate/data-model';
+import { isLdcOrDiscontinuanceOrEnforcementAppealType } from './appeal-type-checks.js';
+
+/**
+ * @param {string | undefined} appealType
+ * @param {string | undefined} procedureType
+ * @returns {boolean}
+ */
+export const displayFinalComments = (appealType, procedureType) =>
```
topickering, line 9: Let's hope that the business-rules file expands!

topickering (approved):

PR #3908, opened 2026-07-13
feat(appeals): create individual call for appeal details page to get only required data(A2-8260)

lachlandow1:
## Describe your changes

Updates:
Dedicated API Endpoint: Added new route GET /appeals/:appealId/page-details specifically tailored to the appeal details page.
This query gets the specific data required for the appeal details page to load. Reducing document retrieval to 1 document and getting the document count for the page. Getting only the latest document for when the data is required.
Added document count for the document page on the appeal-details page.

Removed additional round trip Queries and document payload and made into one single response/

WEB updates:
Updated appeal-details.middleware.js and appeal-details.service.js to call getAppealDetailsPageFromId.
Updated associated mappers (status-tags.mapper.js, display-page-formatter.js, environmental-assessment.mapper.js) to handle document count metadata.

Testing:
Updated unit test suites across API and Web packages to align with new setup

Metrics:

Metric | Legacy Endpoint (/appeals/:appealId) | Optimized Endpoint (/appeals/:appealId/page-details) | Improvement / Impact
-- | -- | -- | --
Response Latency | ~9.49 ms | ~1.84 ms | ~80% faster (~5x latency reduction)

[Ticket](https://pins-ds.atlassian.net/browse/A2-8260)

### DIFF
--- a/appeals/api/src/server/repositories/appeal.repository.js
+++ b/appeals/api/src/server/repositories/appeal.repository.js
@@ -1076,11 +1076,381 @@ const getAppealReference = async (appealId) => {
 	return appeal?.reference ?? '';
 };
 
+/**
+ * @type {import('#db-client/models.ts').AppealSelect}
+ **/
+export const appealDetailsPageDisplaySelect = /** @type {Object} */ {
+	id: true,
+	reference: true,
+	currentStatus: true,
+	applicationReference: true,
+	caseCreatedDate: true,
+	caseValidDate: true,
+	caseExtensionDate: true,
+	caseStartedDate: true,
+	withdrawalRequestDate: true,
+	caseResubmittedTypeId: true,
+	caseTransferredId: true,
+	eiaScreeningRequired: true,
+	padsInspectorUserId: true,
+	address: true,
+	procedureType: {
+		select: {
+			name: true
+		}
+	},
+	parentAppeals: {
+		select: {
+			type: true,
+			linkingDate: true,
+			parent: {
+				select: {
+					id: true,
+					reference: true
+				}
+			}
+		}
+	},
+	childAppeals: {
+		select: {
+			type: true,
+			linkingDate: true,
+			child: {
+				select: {
+					id: true,
+					reference: true
+				}
+			}
+		}
+	},
+	neighbouringSites: {
+		select: {
+			id: true,
+			source: true,
+			address: {
+				select: {
+					addressLine1: true,
+					addressLine2: true,
+					addressTown: true,
+					addressCounty: true,
+					postcode: true
+				}
+			}
+		}
+	},
+	allocation: {
+		select: {
+			level: true,
+			band: true
+		}
+	},
+	specialisms: {
+		select: {
+			specialism: {
+				select: {
+					name: true
+				}
+			}
+		}
+	},
+	appellantCase: {
+		select: {
+			id: true,
+			appellantCaseValidationOutcome: true,
+			knowsOtherOwners: true,
+			knowsAllOwners: true,
+			appellantCaseAdvertDetails: true,
+			contactAddress: true,
+			siteSafetyDetails: true,
+			numberOfResidencesNetChange: true,
+			screeningOpinionIndicatesEiaRequired: true,
+			applicationMadeUnderActSection: true,
+			planningObligation: true,
+			statusPlanningObligation: true,
+			enforcementReference: true
+		}
+	},
+	appellant: {
+		select: {
+			id: true,
+			firstName: true,
+			lastName: true,
+			organisationName: true,
+			email: true,
+			phoneNumber: true
+		}
+	},
+	agent: {
+		select: {
+			id: true,
+			firstName: true,
+			lastName: true,
+			organisationName: true,
+			email: true,
+			phoneNumber: true
+		}
+	},
+	lpa: {
+		select: {
+			name: true,
+			email: true
+		}
+	},
+	appealStatus: {
+		select: {
+			status: true,
+			valid: true,
+			createdAt: true
+		}
+	},
+	appealTimetable: {
+		select: {
+			id: true,
+			lpaQuestionnaireDueDate: true,
+			caseResubmissionDueDate: true,
+			ipCommentsDueDate: true,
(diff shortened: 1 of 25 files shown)


appeals/web/src/server/appeals/appeal-details/appeal-details.router.js
```diff
@@ -51,15 +56,15 @@ router
...
 	.get(
 		saveBackUrl('appeals-detail'),
-		validateAppeal,
+		validateAppealWithInclude(appealDetailsIncludes),
 		clearSessionData,
 		assertUserHasPermission(
 			permissionNames.viewCaseDetails,
 			permissionNames.viewAssignedCaseDetails
 		),
 		asyncHandler(controller.viewAppealDetails)
```
bryn-pins, line 64: We need to do a lot more than this unfortunately

The page itself does the following additional lookups:

getAppealCaseNotes
getRepresentationsByTypes
getAppellantCaseFromAppealId
getOriginalAndLatestLetterDatesObject
getInvalidStatusCreatedDate
getAppealTypesFromId

These are often inefficient api side as well
e.g.
- getAppealCaseNotes actually looks up casenotes twice in the api. And this could just be retrieved in the original call instead
- getAppellantCaseFromAppealId returns all documents for an appeal

I'm seeing 206 DB queries being made on main, currently brought down to 201 with this change

On a slow high doc count case in staging this is down from from 9 seconds to 8 seconds. But we should be able to get it down to below 1 despite the connection being slow between my machine and staging db.

Probably needs a dedicated API endpoint rather than using the validate appealWithInclude route

lachlandow1: @bryn-pins 
Thanks for the feedback. I've updated this to address those points:

Dedicated Endpoint: Replaced validateAppealWithInclude on the details page with /appeals/:appealId/page-details as a dedicated check for appeal details

Pulled the multiple round trip lookups into the main query for the page

Optimized Document query: Folders now fetch with take: 1 alongside a count for document totals rather then totalling in web.
Testing locally showed a drop ~9.5ms down to ~1.8ms. Let me know if there is anything else/improvements that could be done

appeals/api/src/server/repositories/appeal.repository.js
```diff
@@ -1091,11 +1091,271 @@ const getAppealReference = async (appealId) => {
...
+ **/
+export const appealDetailsPageDisplayInclude = /** @type {Object} */ {
+	address: true,
+	procedureType: {
+		select: {
+			id: true,
+			name: true
+		}
+	},
+	parentAppeals: {
```
bryn-pins, line 1102: we call loadLinkedAppeals in loadAndFormatAppeal which fetches this data again I think, perhaps we only need the refs as opposed to the whole linkedAppealsInclude?

something like:
```
include: {
	child: {
		select: {
			id: true,
			reference: true
		}
	}
}
```

appeals/api/src/server/repositories/appeal.repository.js
```diff
@@ -1091,11 +1091,271 @@ const getAppealReference = async (appealId) => {
...
+				}
+			},
+			documents: {
+				where: {
+					isDeleted: false
+				},
+				orderBy: {
+					createdAt: 'desc'
+				},
+				take: 1,
```
bryn-pins, line 1402: nice

appeals/api/src/server/repositories/appeal.repository.js
```diff
@@ -1091,11 +1091,271 @@ const getAppealReference = async (appealId) => {
...
+				select: {
+					guid: true,
+					name: true,
```
bryn-pins, line 1405: Do we need all of this?

Could this just be: 
```
select: {
  guid: true,
  isDeleted: true
}
```

appeals/api/src/server/repositories/appeal.repository.js
```diff
@@ -1091,11 +1091,271 @@ const getAppealReference = async (appealId) => {
...
+	lpa: true,
+	appealStatus: {
+		select: {
+			status: true,
+			valid: true,
+			createdAt: true
+		}
+	},
+	appealTimetable: {
+		select: {
```
bryn-pins, line 1207: Think we need the following adding as well:
```
planningObligationDueDate: true,	statementOfCommonGroundDueDate: true
```

appeals/api/src/server/repositories/appeal.repository.js
```diff
@@ -1091,11 +1091,271 @@ const getAppealReference = async (appealId) => {
...
+							isDeleted: false
+						}
+					}
+				}
+			},
+			documents: {
+				where: {
+					isDeleted: false
+				},
+				orderBy: {
```
bryn-pins, line 1399: can we drop the order by?

bryn-pins (commented):
Down to around 1 second and 50 lookups, big improvement

appeals/api/src/server/repositories/appeal.repository.js
bryn-pins: Or is latest createdAt being used somewhere?

lachlandow1: It's used by the environmental services review for the date received column. We need to keep the orderBy to ensure that we are getting latest document for the mapEnvironmentalAssessment function - I've removed some more of the fields from the selected that aren't essential

appeals/api/src/server/repositories/appeal.repository.js
lachlandow1: updated

appeals/api/src/server/repositories/appeal.repository.js
lachlandow1: Yes, I've updated for this

appeals/api/src/server/repositories/appeal.repository.js
lachlandow1: Reviewed all the included fields for for this and removed where possible

lachlandow1:
Updated to use a select statement for the appeal details call

bryn-pins (approved):

PR #3914, opened 2026-07-15
fix(appeals): site visit transition fixes (A2-8796)

lachlandow1:
## Describe your changes

Site Visit Transitions: Restricted the site visit scheduling/updating state transition so that appeals (and their linked child appeals) only transition to Awaiting Site Visit (i.e. awaiting_event) if their current status is event. Updating a site visit on an appeal in any other state (e.g. final_comments) will no longer trigger an automatic status transition.

Linked Appeal Support: Updated the state transition logic match these status changes to linked child appeals if they match the parent's status.

Tests:
Added new tests transition-state.test.js to test transition logic from final comments to event, awaiting event and issue decision.
Added unit tests in site-visits-update.test.js covering linked appeal status transitions during site visit updates (verifying correct behavior for matching statuses, differing child statuses, non-event statuses, and ensuring related non-linked appeals are not updated).

[Ticket](https://pins-ds.atlassian.net/browse/A2-8796)

### DIFF
--- a/appeals/api/src/server/endpoints/site-visits/__tests__/site-visits-update.test.js
+++ b/appeals/api/src/server/endpoints/site-visits/__tests__/site-visits-update.test.js
@@ -2,6 +2,8 @@
 import { jest } from '@jest/globals';
 import {
 	AUDIT_TRAIL_SITE_VISIT_TYPE_SELECTED,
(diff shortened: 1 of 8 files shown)


DavidJWhiteDev (approved):
lgtm

PR #3915, opened 2026-07-15
feat(web): added expedite questions to LPAQ CAS planning (A2-8700)

musabadam-pins:
## Describe your changes

- Added AnySignificantChangesLPA question
- Added listOfDocumentsBeforeDecision question
- Added original evidence section and questions
- decoupled HAS and CAS Planning mappings to make it easier to do type specific changes
- Updated CAS Planning PDF mapping

[Ticket](https://pins-ds.atlassian.net/browse/A2-8700)

### DIFF
--- a/appeals/pdf-service-api/src/mappers/lpa-questionnaire/sections/original-evidence/row-keys.js
+++ b/appeals/pdf-service-api/src/mappers/lpa-questionnaire/sections/original-evidence/row-keys.js
@@ -34,6 +34,36 @@ export const rowKeys = {
 		'additionalDocumentsLPA',
 		'listOfDocumentsBeforeDecision'
 	],
+	[APPEAL_TYPE.CAS_PLANNING]: [
+		{
+			key: 'didSubmitDesignAccessStatementLPA',
+			condition: (data) => !beforeExpeditedOriginalApplicationCutOff(data.applicationDate)
+		},
+		{
+			key: 'designAccessStatementLPA',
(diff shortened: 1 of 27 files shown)


appeals/api/src/server/mappers/integration/commands/index.js
```diff
@@ -119,7 +119,7 @@ const mapAppealSubmission = (data) => {
 const mapQuestionnaireSubmission = (data, appeal, designatedSites) => {
 	const { casedata, documents } = data;
 	const questionnaireInput = {
-		...mapQuestionnaireIn({ casedata }, designatedSites),
+		...mapQuestionnaireIn({ casedata }, designatedSites, appeal?.caseCreatedDate || null),
```
lachlandow1: Should this be the applicationDate from appeal case? can't remember how this is  mapping so just confirming that this is the right date

appeals/pdf-service-api/src/mappers/lpa-questionnaire/sections/original-evidence/row-keys.js
lachlandow1, line 1: same here - we have used data.applicationDate in the past

lachlandow1:
Just generally confirming that we are using the appellants answer to the application date question to determine the expedited eligible logic

appeals/api/src/server/mappers/integration/commands/index.js
musabadam-pins: you are correct, its been changed in the CAS advert version of this ticket

lachlandow1 (approved):

PR #3920, opened 2026-07-16
fix(api): allowing spaces in filename regex (a2-8791)

pinsMT:
- Updated regex

Ticket: https://pins-ds.atlassian.net/browse/A2-8791

### DIFF
--- a/appeals/api/src/server/common/validators/filename-parameter.js
+++ b/appeals/api/src/server/common/validators/filename-parameter.js
@@ -11,5 +11,5 @@ export const validateFileNameParameter = (parameterName) =>
 	body(parameterName)
 		.optional({ checkFalsy: true })
 		// Filename must only contain alphanumeric characters, underscores, hyphens and one period followed by a suffix
-		.matches('^[a-zA-Z0-9_-]+\\.[a-zA-Z0-9_-]+$')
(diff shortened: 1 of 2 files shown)


appeals/api/src/server/common/validators/filename-parameter.js
```diff
@@ -11,5 +11,5 @@ export const validateFileNameParameter = (parameterName) =>
 	body(parameterName)
 		.optional({ checkFalsy: true })
 		// Filename must only contain alphanumeric characters, underscores, hyphens and one period followed by a suffix
-		.matches('^[a-zA-Z0-9_-]+\\.[a-zA-Z0-9_-]+$')
+		.matches('^[a-zA-Z0-9_ -]+\\.[a-zA-Z0-9_-]+$')
```
zoebutterworth, line 14: Do you need to accept spaces after the \ as well?

zoebutterworth (dismissed):

appeals/api/src/server/common/validators/filename-parameter.js
pinsMT: determined this is not nescessary

zoebutterworth (approved):

PR #3927, opened 2026-07-20
fix(web): CAS Adverts expedite LPAQ updates (A2-8701)

musabadam-pins:
## Describe your changes

- Added AnySignificantChangesLPA question
- Added listOfDocumentsBeforeDecision question
- Added original evidence section and questions
- Updated import and broadcast mappings
- Added applicationDate under applicatantSubmission when grabbing appeal data
- Updated PDF download 
- Updated api that calls PDF download and pass apellantSubmission with LPAQ as applicationDate attribute is needed for expedite check

[Ticket](https://pins-ds.atlassian.net/browse/A2-8701)

### DIFF
--- a/appeals/api/src/server/mappers/mapper-factory.js
+++ b/appeals/api/src/server/mappers/mapper-factory.js
@@ -1,7 +1,10 @@
 import { isFeatureActive } from '#utils/feature-flags.js';
 import mergeMaps from '#utils/merge-maps.js';
 import { FEATURE_FLAG_NAMES } from '@pins/appeals/constants/common.js';
-import { isS78ExpeditedAppealType } from '@pins/appeals/utils/appeal-type-checks.js';
+import {
+	beforeExpeditedOriginalApplicationCutOff,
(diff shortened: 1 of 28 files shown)


pinsMT (approved):

appeals/api/src/server/mappers/mapper-factory.js
```diff
@@ -79,8 +82,13 @@ function createDataMap(mappingRequest) {
 			}
 		}
 		case APPEAL_CASE_TYPE.ZA: {
-			const casAdvert = createMap(apiMappers.apiCasAdvertMappers, mappingRequest);
-			return mergeMaps(caseData, casAdvert);
+			if (!beforeExpeditedOriginalApplicationCutOff(appeal.appellantCase?.applicationDate)) {
```
lachlandow1, line 85: Not sure about the logic here - if the application date is after 1st april the old mappers are used while if before used the expedited - should this be the other way round?

lachlandow1 (approved):

PR #3931, opened 2026-07-20
fix(api): allowing brackets in doc filename validation (a2-8825)

pinsMT:
- Added brackets to api-side filename validator regex
- Added brackets to web-side validator
- Added test for checking brackets web & api
- Updated banner error message to include new conditions
- Updated tests to reflect changes

Ticket: https://pins-ds.atlassian.net/browse/A2-8825


DavidJWhiteDev (approved):
nice one, tests aok!

PR #3950, opened 2026-07-23
test(appeals): add tests for filename validation (A2-8825)

marcpodgorney81:
## Describe your changes

PR for adding tests which check that valid and invalid filename formats are handled as expected 
This is following on from discovery that spaces and brackets (both allowed) were being processes as invalid 

Added them to the globalValidation.spec file, as seemed logical, given are multiple places in backoffice where file can be renamed? 

Also have add a couple of tests to the manageDocsAppellant.spec to check that can upload files with spaces and brackets in the name and rename them (another use case that was found)    

Other changes 
- have updated globalValidation to combine date and time, each set of tests now has it's own before and after 
- added some more properties to consts.js file for example filenames, messages e.g. 
- added FileDetailsPage to represent file details screen  

Tickets 
https://pins-ds.atlassian.net/browse/A2-8825 
https://pins-ds.atlassian.net/browse/A2-8818


jacksolirius (dismissed):
LGTM

jacksolirius (approved):
LGTM

PR #3966, opened 2026-07-28
A2 7894 Adding tests for expedited appeals 

marcpodgorney81:
## Describe your changes

This PR adds tests for the expedited appeals process 

**Carries out following flow:**   
- submit appeal 
- Validate appeal 
- Check available appeal types on choose procedure screen 
- Start case 
- Check details in case overview 
- Check appeals data from (i.e. after fetching from DB) 

And does this for both expedited and non expedited appeals 

**At the moment is running on the following appeal types**   
- Full planning 
- Outline planning 
- Reserved matters 
- Prior approval 

Should be fairly straightforward to add in additional appeal types, e.g. remove/vary condition and permission in principle when implemented 

Also renamed file to expeditedAppeals.spec as felt more descriptive 

**Some remaining issues/todos**  
- Think test payload still needs some tweaking 
- Need to add start case check to remaining scenarios 
- Potentially add some more scenarios  
- Might be good to add something for HAS/CAS? 
- Add scenario for environmental appeal 

Might be good as well (as follow up) to add tests for e2e of expedited appeal (i.e. up to issue decision)? 

Ticket https://pins-ds.atlassian.net/browse/A2-7894


hardiebrown1 (approved):
Nice work @marcpodgorney81 Great foundation built for expedited going forward.

I agree with your next steps/ areas we can develop further. Also think we should have some validation for Appellant form questions and LPAQ questions.

marcpodgorney81:
> Nice work @marcpodgorney81 Great foundation built for expedited going forward.
> 
> I agree with your next steps/ areas we can develop further. Also think we should have some validation for Appellant form questions and LPAQ questions.

Thanks @hardiebrown1! 
Yes, think I meant to add something to that next steps list about appellant form!

PR #3969, opened 2026-07-29
fix(web): removed significantchanges from s78 shared (A2-8891)

musabadam-pins:

[Ticket]()


zoebutterworth (approved):
I agree this matches with what was merged into main :)

lachlandow1 (approved):

PR #3973, opened 2026-07-29
chore(appeals): notifications and triggers document update (no-ticket)

zoebutterworth:
## Describe your changes

- Adding most templates to the docs (it doesn't include all the enforcement templates which exist)
- Adding links to the subjects as well as the content
- Adding GOV.UK notify links where they are known

No ticket

### DIFF
--- a/docs/notifications-and-triggers.md
+++ b/docs/notifications-and-triggers.md
@@ -8,22 +8,64 @@ these [Notify Templates](../appeals/api/src/server/notify/templates):
 ### Appeal type change
 
 - **Appeal type:** all
-- **Notify Template:** [appeal-type-change-non-has](../appeals/api/src/server/notify/templates/appeal-type-change-non-has.content.md)
+- **Notify Subject Template:** [appeal-type-change-non-has](../appeals/api/src/server/notify/templates/appeal-type-change-non-has.subject.md)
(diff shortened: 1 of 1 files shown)


docs/notifications-and-triggers.md
```diff
@@ -8,185 +8,271 @@ these [Notify Templates](../appeals/api/src/server/notify/templates):
...
+- **Trigger:**
+
+### Appeal type change in MA - lpa
+
+- **Appeal type:**
+- **Notify Subject Template:** [appeal-type-change-in-manage-appeals-lpa](../appeals/api/src/server/notify/templates/appeal-type-change-in-manage-appeals-lpa.subject.md)
+- **Notify Content Template:** [appeal-type-change-in-manage-appeals-lpa](../appeals/api/src/server/notify/templates/appeal-type-change-in-manage-appeals-lpa.content.md)
+- **Trigger:**
+
+## Change procedure type
```
musabadam-pins, line 30: dont have much knowledge on this but is this correct with two lines containing same content?

musabadam-pins (approved):

docs/notifications-and-triggers.md
zoebutterworth: It is just because of the headers, so the first one is a H2 which we use for the sections and the second one is a H3 which we use for the template titles. The different levels of header are indicated by having more (or less) #s

PR #3984, opened 2026-08-03
feat(appeals): adding notify docs check to commit verify (no-ticket)

zoebutterworth:
## Describe your changes

- Adding a check to the commit verify that all the notify templates are included in the notifications-and-triggers doc

Note that this doesn't check that the list in the doc exactly matches the templates and so there may be templates in the doc which no longer exist in the folder. It also doesn't examine the content- just that the name of the template is in the document. These could be added in the future as improvements :)

No ticket


bryn-pins (approved):
Seems sensible

PR #3986, opened 2026-08-03
fix(appeals): adding specific case history for new expedited appellan…

zoebutterworth:
…t case question (A2-8025)

and updating the generic appellant case audit message

## Describe your changes

- Adding the specific case history message for 'Why are you appealing?'
- Updating the generic case history message for appellant case to 'Appellant case updated' in order to match with 'LPA questionnaire updated'
- Updated tests for new wording
- Added tests for the two questions in the ticket

### Tested locally

image

[Ticket](https://pins-ds.atlassian.net/browse/A2-8025)

### DIFF
--- a/appeals/api/src/server/endpoints/appellant-cases/appellant-cases.service.js
+++ b/appeals/api/src/server/endpoints/appellant-cases/appellant-cases.service.js
@@ -801,7 +801,8 @@ export function renderAuditTrailDetail(data) {
 				/** @type {string} */ (data.applicationMadeUnderActSection || '').replaceAll('-', ' ')
 			),
 		AUDIT_TRAIL_SCREENING_OPINION_INDICATES_EIA_REQUIRED_UPDATED: () =>
-			data.screeningOpinionIndicatesEiaRequired ? 'Yes' : 'No'
+			data.screeningOpinionIndicatesEiaRequired ? 'Yes' : 'No',
+		AUDIT_TRAIL_REASON_FOR_APPEAL_APPELLANT_UPDATED: () => data.reasonForAppealAppellant
 	};
 
 	if (!auditTrailParameters[constantKey]) {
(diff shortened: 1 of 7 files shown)


appeals/api/src/server/endpoints/appellant-cases/appellant-cases.service.js
```diff
@@ -801,7 +801,8 @@ export function renderAuditTrailDetail(data) {
 				/** @type {string} */ (data.applicationMadeUnderActSection || '').replaceAll('-', ' ')
 			),
 		AUDIT_TRAIL_SCREENING_OPINION_INDICATES_EIA_REQUIRED_UPDATED: () =>
-			data.screeningOpinionIndicatesEiaRequired ? 'Yes' : 'No'
+			data.screeningOpinionIndicatesEiaRequired ? 'Yes' : 'No',
+		AUDIT_TRAIL_REASON_FOR_APPEAL_APPELLANT_UPDATED: () => data.reasonForAppealAppellant
```
bryn-pins, line 805: No issue with this being nvarchar max
Could be huge on the audit history
Perhaps we could trim to a given number of chars either before display or before saving?

bryn-pins (approved):
Just a comment on the potential data length, happy to go in as is, perhaps a separate ticket to resolve for all fields

appeals/api/src/server/endpoints/appellant-cases/appellant-cases.service.js
zoebutterworth: Is this concern due to performance or user experience? It does collapse with a 'read more' button

bryn-pins: Both, but with read more is probably fine removes the user experience issue, probably just need to add paging to the history page at some point in future

PR #3987, opened 2026-08-03
fix(api): common place for appeal detail keys, that test util reads f…

sam-laskowski:
…rom (A2-8487)

## Describe your changes

- add in procedureType: true to prisma include

- Tested by running /appeals/{appealReference}/share-comments-and-statement locally, against an s20 Hearing appeal, sends correct notify.

[Hearing -Incorrect notifies are sending out to Appellant & LPA upon sharing statements & IP comments.](https://pins-ds.atlassian.net.mcas.ms/browse/A2-8487)

### DIFF
--- a/appeals/api/src/server/endpoints/test-utils/test-utils.controller.js
+++ b/appeals/api/src/server/endpoints/test-utils/test-utils.controller.js
@@ -453,7 +453,7 @@ export const simulateShareIpCommentsAndLpaStatement = async (req, res) => {
 	const { appealReference } = req.params;
 	const appeal = await databaseConnector.appeal.findUnique({
 		where: { reference: appealReference },
-		include: { appealStatus: true, lpa: true, appellant: true, agent: true }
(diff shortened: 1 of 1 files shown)


appeals/api/src/server/common/appeal-keys/selected-keys.js
```diff
@@ -0,0 +1,23 @@
+/** @type {Array<keyof typeof import('#repositories/appeal.repository.js').appealDetailsIncludeMap>} */
+export const selectedKeys = [
```
zoebutterworth: I'm not sure about pulling this out into a common file like this- selectedKeys is not a very clear name for this and the 'selectedKeys' are different on pretty much every route we have

sam-laskowski: I've made a change to just add in procedureType to keep everything consistent with its current state, retested to check that it still fixes bug.

zoebutterworth (approved):

PR #4000, opened 2026-08-06
fix(web): remove display for unneeded due dates for writtenPart1 procedure type (A2-8946)

lachlandow1:
### Updates:
Remove the final comments row for writtenPart1 procedure type.
Prevent generation of unneeded due dates for S78 expedited
remove check of s78 expedited appeal type as it doesn't cover the display of rows as required

### Tests:
Added unit test for the row display of S78 expedited appeals.
Manually tested in browser

[Ticket](https://pins-ds.atlassian.net/browse/A2-8946)

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/__tests__/appeal-details.test.js
+++ b/appeals/web/src/server/appeals/appeal-details/__tests__/appeal-details.test.js
@@ -5301,6 +5301,37 @@ describe('appeal-details', () => {
 					expect(caseTimeTable).toMatchSnapshot();
 				});
 
+				it('should not render LPA statement, IP comments, Statement of common ground, or Final comments due dates for S78 Part 1 procedureType appeals', async () => {
+					nock('http://test/')
+						.get(`/appeals/${appealId}/page-details`)
+						.reply(200, {
+							...appealDataFullPlanning,
+							caseOfficer: '2cb7735e-c4cf-410b-b773-5ec4cf110b87',
+							appealId,
+							procedureType: 'Part 1',
+							appealTimetable: {
+								lpaQuestionnaireDueDate: '2026-04-10T23:59:00.000Z',
+								lpaStatementDueDate: '2026-04-20T23:59:00.000Z',
+								ipCommentsDueDate: '2026-04-20T23:59:00.000Z',
+								statementOfCommonGroundDueDate: '2026-04-20T23:59:00.000Z',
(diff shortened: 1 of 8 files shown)


bryn-pins (dismissed):

appeals/web/src/server/appeals/appeal-details/appeal-details-page-sections/s78.js
```diff
@@ -50,13 +53,21 @@ export function generateAppealDetailsPageComponents(appealDetails, mappedData, s
...
+								!isWrittenPart1
+									? mappedData.appeal.statementOfCommonGroundDueDate.display.summaryListItem
+									: undefined,
 								mappedData.appeal.planningObligationDueDate.display.summaryListItem,
 								mappedData.appeal.proofOfEvidenceAndWitnessesDueDate.display.summaryListItem,
 								mappedData.appeal.caseManagementConferenceDueDate.display.summaryListItem,
-								mappedData.appeal.finalCommentDueDate.display.summaryListItem,
+								!isWrittenPart1
+									? mappedData.appeal.finalCommentDueDate.display.summaryListItem
+									: undefined,
```
zoebutterworth: I'm not a huge fan of having some of the logic around the rows displaying in the mappers and some here. Was there a specific reason for having it here rather than in the mapper files?

lachlandow1: Was looking at centralising the control for this. But yes probably better off in the mappers only. Making the updates

appeals/web/src/server/lib/mappers/data/appeal/submappers/final-comment-due-date.mapper.js
```diff
@@ -17,6 +18,8 @@ export const mapFinalCommentDueDate = ({
 	// otherwise we only have final comments for written procedure type
 	if (
 		!appealDetails.startedAt ||
+		appealDetails.procedureType === APPEAL_CASE_PROCEDURE.WRITTEN_PART_1 ||
+		appealDetails.procedureType === PROCEDURE_TYPE_NAME.WRITTEN_PART_1 ||
```
zoebutterworth: As I assume these don't ever want to display anything final comments can they be added as a criteria in displayFinalComments as called below? Ideally in the long term all of these deadlines will have functions in business-rules and then we will change things centrally there but that isn't necessarily something for this piece of work

zoebutterworth (approved):

PR #4020, opened 2026-08-12
feat(api): removed files, added new files which handle multiple statements results (a2-9003)

pinsMT:
- Added new files using conditions to map out statements/comments list and other variables
- Removed files which covered multiple situations handling statements/comments listing
- Changed IF statements to handle selecting new files
- Updated notifications and triggers

Ticket: https://pins-ds.atlassian.net/browse/A2-9002


appeals/api/src/server/notify/templates/publish-statements-enforcement-hearing-no-statements-or-comments-received.subject.md
lachlandow1, line 1: We should be able to cover this in [publish-statements-enforcement-hearing-statements-or-comments-received.content.md](https://github.com/Planning-Inspectorate/appeals-back-office/pull/4020/changes#diff-51a28fee73fd1077213b2ec5c565fd095475a01300f777edef0c5763c4e0fea7)

lachlandow1 (commented):
Great work :)

Since we need to cover all the scenarios that have been deleted. It would be good to have tests to cover all the different template scenarios to increase confidence that everything has been covered. so for each deleted template we should have an expected result for the one template.

Additionally I've added a note to remove the template for no comments or statements to make it that we are only using one template for all scenarios

appeals/api/src/server/notify/templates/publish-statements-enforcement-hearing-no-statements-or-comments-received.subject.md
pinsMT: all sorted in the new commit

lachlandow1 (approved):

